PR #1, opened 2019-08-25
rename TOKEN_MERGE to TOKEN_UPDATE

zimbatm:
according to src/libexpr/lexer.l, nix calls `//` the UPDATE token

see https://github.com/nixos/nix/blob/f435634a29551754d5f7303b0a60cd8fe2df2079/src/libexpr/lexer.l#L122

### DIFF
--- a/src/kinds.rs
+++ b/src/kinds.rs
@@ -37,7 +37,7 @@ pub enum SyntaxKind {
     TOKEN_PAREN_CLOSE,
     TOKEN_CONCAT,
     TOKEN_INVERT,
-    TOKEN_MERGE,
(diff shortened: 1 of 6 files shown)


zimbatm:
There are probably a few more of those but I wanted to start somewhere.

Are you waiting for me to merge the PR or do you want to do it yourself?

zimbatm:
also renamed the Kind to match the token name

jD91mZM2:
I was waiting in case @matklad had any thoughts since you asked him to review too. Will merge.

zimbatm:
it was more to keep him in the loop. see also #2 for a more general ticket to track the differences.

PR #4, opened 2019-09-03
Use consistent unary_op/binary_op naming for operators

matklad:


zimbatm:
In the C++ implementation, the prefix seems to be `ExprOp` for most operations.

For example `==` is translated to a `ExprOpEq($1, $3)`. Or the unary `!` becomes `ExprOpNot($2)`.

Weirdly, some of the operations are translated to builtin function calls directly instead of getting their own AST node type. For example the unary negation becomes `ExprApp(CUR_POS, new ExprApp(new ExprVar(data->symbols.create("__sub")), new ExprInt(0)), $2);`

https://github.com/nixos/nix/blob/f435634a29551754d5f7303b0a60cd8fe2df2079/src/libexpr/parser.y#L342-L363

matklad:
@zimbatm yeah, in genera we won't be able to directly use naming from C++ impl, as that builds an AST, and we are building concrete syntax tree.

PR #6, opened 2019-09-04
rename value -> literal

matklad:
That's pretty subjective as there's no one true terminology, but I
think literal is a better term here:

* value usually refers to the result of evaluation, that is, to a
specific pile of bits in the memory

* literal, on the other hand, usually refers to a bit of syntax

Literal is often, but not always, used to describe "atomic"
self-evaluation constructs like ints or string (for example, in the
Rust reference).

Sometimes it is also used for compound things (ie, "collection
literals"), but that's not the meaning we use here.


zimbatm:
What if TOKEN_VALUE was removed entirely? In all the examples it seems to only contain a single value with the same range. Instead, the TOKEN_INTEGER, TOKEN_FLOAT, ... could share a kind "literal".

matklad:
We distinguish, on the type level, between tokens (leaf nodes of the syntax tree) and nodes (interior nodes of the syntax tree). This is useful because most tokens like `{` or ` ` do not contribute to the semantic level and should be skipped by most tree traversals. 

However, for those tokens that do contribute semantic meaning, we need to create a silly wrapper node which has the token as the sole child, to make the API convenient.

So that's the reason why we want to wrap `TOKEN_INTEGER | TOKEN_FLOAT | etc` into a `NODE_LITERAL` / `NODE_VALUE`.

PR #8, opened 2019-09-04
rename set related nodes

matklad:
Set -> AttrSet
SetEntry -> KeyValue (so it makes more sense inside let)
Attribute -> Key (a dotted key like foo.bar is not exactly an attribute)

Note that upstream uses the term "bind"` instead of `key_value`.


jD91mZM2:
Thanks for all these PRs! Should I release a new minor version after this? (or even 1.0.0?)

Absolutely awesome work, you're awesome 👍

(note: merge conflict on this final one, i can fix it later today hopefully)

matklad:
fixed the conflict

A new release will be helpful, yeah! Not sure about 1.0 just yet.

PR #13, opened 2020-02-25
Implement more descriptive errors

selaux:
**Goal**:
I had a look at `rnix-lsp` and saw that the error description were not very good, just `error node at 1..2`. In order to support better error messages in `rnix-lsp`, which describe the desired state such as "Unexpected `{` expected `=`" this information would need to be included in the parsing errors.

**Implementation**:
This stores details of error nodes in the parsers errors, which is an approach [also used by rust-analyzer](https://github.com/rust-analyzer/rowan/issues/26). When collecting the error nodes, it checks whether the node is already included in the parsers errors by text position and discards it if so. Ideally only descriptive errors should remain. In order to keep track of the text position i introduced another property on the parser called `index`, which is the offset in bytes the parser is currently at, this is transformed into `TextPositions` in errors.

I also removed the public `root_errors` function. Since errors can now be easily distinguished it did not seem useful to me.


src/parser.rs
```diff
@@ -53,28 +79,34 @@ impl AST {
...
+            .errors
+            .iter()
+            .filter_map(|e| match e {
+                ParseError::UnexpectedWanted(_, t, _) => Some(t),
+                ParseError::UnexpectedDoubleBind(t) => Some(t),
+                ParseError::UnexpectedExtra(t) => Some(t),
+                _ => None,
+            })
+            .collect();
         let mut errors = self.errors.clone();
```
jD91mZM2: Am I correct that this patch supersedes root errors? Because if so, keeping all original errors here may result in duplicates?

src/parser.rs
```diff
@@ -295,9 +348,13 @@ where
         if self.peek() == Some(TOKEN_AT) {
             let kind = if bound { NODE_ERROR } else { NODE_PAT_BIND };
             self.start_node(kind);
+            let start = self.get_text_position();
             self.bump();
             self.expect_ident();
-            self.finish_node();
+            let end = self.finish_error_node();
+            if kind == NODE_ERROR {
```
jD91mZM2: We probably want to use `if bound` here :)

src/parser.rs
```diff
@@ -455,10 +512,25 @@ where
...
+            kind => {
+                let start = self.start_error_node();
                 self.bump();
-                self.finish_node();
+                let end = self.finish_error_node();
+                self.errors.push(ParseError::UnexpectedWanted(
+                    kind,
+                    TextRange::from_to(start, end),
+                    [
+                        TOKEN_PAREN_OPEN,
```
jD91mZM2: Do you think this is warranted a helper function?

src/parser.rs
selaux: The error nodes that we collect from the tree should now be a superset of `self.errors`, basically containing the more detailed variant of the error node. This is why we filter `self.root().errors()` below. Ideally all of the errors in `self.root().errors()` should also be in self.errors, but right now there is no guarantee. Long story short: the filter below should exclude duplicates.

src/parser.rs
selaux: Will change.

src/parser.rs
selaux: I dont know if this makes sense here as you would need to pass the list of expected tokens anyways, thus it will not safe a lot of space, or make it more readable imho.

selaux:
One more thing I noticed: I still need to adapt the example that prints errors. It only checks for `ParserError::Unexpected` pattern. This might generally be an issue for consumers of the library, at it is not really clear now that there are more variants to `ParserError` than before. So people having a `_ =>` match arm will not notice that there are any changes. What do you think? Does this need to be taken care of?

src/parser.rs
selaux: Fixed.

selaux:
The example is fixed as well.

jD91mZM2:
Yeah, don't mind the consumers, we're not 1.0.0 :D

PR #18, opened 2020-03-15
Implement TypedNode for ParsedType

selaux:
This allows `ParsedType`, which only encapsulates items with trait `TypedNode` to be used as a `TypedNode`.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -1,4 +1,5 @@
 //! Provides a type system for the AST, in some sense
+use std::convert::TryFrom;
 use std::fmt;
 
 use crate::{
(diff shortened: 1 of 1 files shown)


zimbatm:
what is the motivation behind this change?

selaux:
The motivation is that I want to have easy access to the node the `ParsedType` encapsulates. E.g. I have some instance of `ParsedType` such as `ParsedType::Ident()` and want to access the underlying `SyntaxNode`. The only way to do this right now (that I know of) is to write a huge match expression (basically the one included in the PR). But from consumer perspective I want to have as little friction as possible in coverting from typed to untyped nodes and vice versa. Does that make sense?

PR #21, opened 2020-03-29
Support parsing dynamic inherit identifiers

andir:
Nix allows code like `inherit ${"foo"}` to inherit the "foo" attribute.
This was previously parsed as invalid code while nix allows using that.

This is a follow-up to a conversation I had with @jD91mZM2 a while ago on IRC. I hope I got the implementation right because I mostly guessed from the notes I took back then.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -374,8 +374,8 @@ where
                         self.finish_node();
                     }
 
-                    while let Some(TOKEN_IDENT) = self.peek() {
-                        self.expect_ident();
+                    while self.peek() != Some(TOKEN_SEMICOLON) {
+                        self.parse_val();
                     }
 
                     self.expect(TOKEN_SEMICOLON);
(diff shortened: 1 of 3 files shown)


src/parser.rs
```diff
@@ -378,6 +378,10 @@ where
                         self.expect_ident();
```
jD91mZM2: This can now go, it's redundant to first parse only idents and then anything :)

src/parser.rs
```diff
@@ -378,6 +378,10 @@ where
...
                     }
 
+                    while self.peek() != Some(TOKEN_SEMICOLON) {
+                        self.parse_expr();
```
jD91mZM2: I believe I tricked you into writing this, I forgot `parse_expr` handles functions and we don't want `inherit a b` to be a function call. Changing it to just `parse_val` should work, though :)

andir:
@jD91mZM2 all requested changes should have been addressed

Issue #33, opened 2021-09-25
Error handling for Select expressions


aaronjanse:
The following is a common scenario when editing Nix:
```nix
rec {
    foo.bar = 13;
    xyz = foo.; # trailing period was just typed
}
```

`rnix-parser` consumes the rest of the file as an error, rather than stopping at the semicolon:

<details>
<summary>`dump-ast` output</summary>
<pre>
error: unexpected TOKEN_SEMICOLON at 38..41, wanted any of [TOKEN_IDENT]
error: unexpected end of file, wanted any of [TOKEN_IDENT]
error: unexpected end of file, wanted any of [TOKEN_SEMICOLON]
error: unexpected end of file
NODE_ROOT 0..41 {
  NODE_ATTR_SET 0..41 {
    TOKEN_REC("rec") 0..3
    TOKEN_WHITESPACE(" ") 3..4
    TOKEN_CURLY_B_OPEN("{") 4..5
    TOKEN_WHITESPACE("\n    ") 5..10
    NODE_KEY_VALUE 10..23 {
      NODE_KEY 10..17 {
        NODE_IDENT 10..13 {
          TOKEN_IDENT("foo") 10..13
        }
        TOKEN_DOT(".") 13..14
        NODE_IDENT 14..17 {
          TOKEN_IDENT("bar") 14..17
        }
      }
      TOKEN_WHITESPACE(" ") 17..18
      TOKEN_ASSIGN("=") 18..19
      TOKEN_WHITESPACE(" ") 19..20
      NODE_LITERAL 20..22 {
        TOKEN_INTEGER("13") 20..22
      }
      TOKEN_SEMICOLON(";") 22..23
    }
    TOKEN_WHITESPACE("\n    ") 23..28
    NODE_KEY_VALUE 28..41 {
      NODE_KEY 28..31 {
        NODE_IDENT 28..31 {
          TOKEN_IDENT("xyz") 28..31
        }
      }
      TOKEN_WHITESPACE(" ") 31..32
      TOKEN_ASSIGN("=") 32..33
      TOKEN_WHITESPACE(" ") 33..34
      NODE_SELECT 34..41 {
        NODE_IDENT 34..37 {
          TOKEN_IDENT("foo") 34..37
        }
        TOKEN_DOT(".") 37..38
        NODE_ERROR 38..41 {
          TOKEN_SEMICOLON(";") 38..39
          TOKEN_WHITESPACE("\n") 39..40
[...]

PR #34, opened 2021-09-26
handle syntax errors more gracefully

aaronjanse:
Fixes https://github.com/nix-community/rnix-parser/issues/33.

I don't know enough about `rowan` to cleanly improve error handling, but I thought I'd at least share the patch I've been using locally while implementing rnix-lsp autocomplete.

Feel free to re-use parts of this patch for a different PR, push directly to this PR, or something else.


src/parser.rs
```diff
@@ -291,7 +301,15 @@ where
...
             Some(TOKEN_DYNAMIC_START) => self.parse_dynamic(),
             Some(TOKEN_STRING_START) => self.parse_string(),
-            _ => self.expect_ident(),
+            Some(TOKEN_IDENT) => self.expect_ident(),
+            _ => {
+                let start = self.start_error_node();
+                let end = self.finish_error_node();
+                self.errors.push(ParseError::Expected(
+                    TextRange::new(start, end),
+                    "attribute".to_string(),
```
Ma27, line 310: I guess we should reflect here that strings etc. are also OK.

Ma27 (commented):
LGTM mostly, nice work! :) 

Would you mind resolving the last comment and rebase?

src/parser.rs
aaronjanse: What should be the exact message used? Maybe `attribute or string`? Although I guess a string does represent an attribute here? Please correct me if I'm wrong

aaronjanse:
I pushed a somewhat significant change to behavior. The change matters only for attr path now. I also added an "expected semicolon" error node and updated tests

oberblastmeister:
You need to run this with a fuzzer. Adding better error handling can easily screw up the parser, and I got an out of memory error relatively quickly. I would prefer more general error handling (removing `expect_peek_any`). It would be also be good to see the effectiveness of the error handling by hooking it up with `rnix-lsp`. Maybe a mono repo would be good for this?

aaronjanse:
Wow, good catch @oberblastmeister. The fuzzer is amazing, and after seeings its result, I agree that we need a more general  solution here. I'll start working on that. Thank you!!

oberblastmeister:
@aaronjanse I already did a lot of work here https://github.com/nix-community/rnix-parser/pull/35. Also, I really think we should work on hooking up the new version of `rnix` to `rnix-lsp` before we do improved error handling. This way we can try out the error handling in real time.

aaronjanse:
> @aaronjanse I already did a lot of work here https://github.com/nix-community/rnix-parser/pull/35

Oh, sorry about that! I'm still catching up after being away for a while. If you submit a PR for the error handling that was in #35, I'll close #120 in favor of your PR.

> I really think we should work on hooking up the new version of rnix to rnix-lsp before we do improved error handling. This way we can try out the error handling in real time.

As in, you think we should update rnix-lsp to use the latest nix-parser version? I'd be willing to work on doing that

oberblastmeister:
> As in, you think we should update rnix-lsp to use the latest nix-parser version? I'd be willing to work on doing that

Yes! I think that would be great.

aaronjanse:
Sounds good, will do!

Issue #32, opened 2021-09-19
Try out fuzzing on the parser


Ma27:
#31 has shown that especially error conditions are interesting and potentially error-prone and there's a certain chance that there are a few more issues like this.

While talking about this with @lheckemann today, he suggested to try out fuzzing, e.g. https://afl-1.readthedocs.io/en/latest/fuzzing.html and I think that's worth a shot :)

Ma27:
Corresponding PR is part of 0.11.0.

PR #35, opened 2021-09-28
[WIP] parsing fixes, better error handling, fuzzing

oberblastmeister:
Currently the parser doesn't parse operator associations correctly, everything is left associative. This is wrong for operators like `++`, `->`, and `//`. I fixed it for those operators. I also incorporated #34. But this error handling is more general and not just for semicolons. It uses the concept of recovery which is taken from rust-analyzer. I added fuzzing for the lexer and the parser which addresses #32. I have fixed the parser to pass all the fuzz tests. I removed `cbitset` which seems like an unnecessary dependency because we don't need all the different bitset variants, only the largest. It is easy just to write one by hand and it is very short. I have put the bitset into use everywhere to check if a token is contained inside rather than using a slice which has to be looped over. My code is very messy right now and I don't have tests, but I will add them as I go.


aaronjanse:
Whoa, this is very cool. Thank you @oberblastmeister!! I'll start trying this out locally while developing rnix-lsp :-)

fuzz/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+name = "parser"
+path = "fuzz_targets/parser.rs"
+test = false
+doc = false
+
+[[bin]]
+name = "lexer"
+path = "fuzz_targets/lexer.rs"
+test = false
+doc = false
```
SuperSandro2000, line 31: Missing final new line

fuzz/fuzz_targets/parser.rs
```diff
@@ -0,0 +1,14 @@
...
+
+fuzz_target!(|data: &[u8]| {
+    let stdout = io::stdout();
+    let mut handle = stdout.lock();
+
+    if let Ok(text) = std::str::from_utf8(data) {
+        writeln!(handle, "Fuzzing {:?}\n\n", data).unwrap();
+        let _ = rnix::parse(text);
+    }
+});
```
SuperSandro2000, line 14: Missing final new line

Ma27:
So this is actually way too much to merge as-is IMHO.

I'm strongly in favor of applying the fuzzer-part, but I'm not sure about the rest currently.

oberblastmeister:
Well I think that our parser actually being able to parse stuff correctly is pretty important :].

aaronjanse:
Yep I agree, correct parsing is important. And improved error handling is necessary for providing autocomplete for rnix-lsp.

@oberblastmeister Would you mind splitting this into smaller PRs, such as maybe:
- fuzzing
- error handling
- correct operation handling
- dependency changes

oberblastmeister:
Yeah I can split them up

aaronjanse:
Thank you!!

Ma27:
> Well I think that our parser actually being able to parse stuff correctly is pretty important :].

I have to apologize actually, my wording was rather poorly! What I meant was exactly what @aaronjanse meant (we briefly talked about it previously), so thanks a lot for your work on this, but splitting things up would actually help us to get things in! :)

mohe2015:
> It uses the concept of recovery which is taken from rust-analyzer.

Does this mean the parser (I don't know the current state) will not be parsing in linear time then? Is the Nix Language not LL(k) parseable?

Ma27:
Re-assigning to 0.11.0. @oberblastmeister do you think you'll be able to get to splitting things up during the next weeks? I'm not sure when I'll get to it, but otherwise I'd resume here :) 

> Does this mean the parser (I don't know the current state) will not be parsing in linear time then? Is the Nix Language not LL(k) parseable?

Perhaps it is, but it doesn't seem desirable. Further context is in https://edolstra.github.io/pubs/phd-thesis.pdf (Section 4.2, page 64):

> Most context-free parsers (e.g., LL(k)
and LR(k)parsers, for fixed k) suffer from bounded look-ahead: they must choose
between different productions on the basis of a fixed number of tokens. This is
inconvenient for language implementors, since they must deal with the resulting
parse conflicts. For instance, the input fragment "{x" could be start of an attribute set
(e.g., {x=123;}), or a function definition (e.g., {x}: x)1. This requires the programmer
to introduce additional production rules to resolve the conflict

oberblastmeister:
This pr should not change the speed of the parser. It does not change what the parser does, only the error handling. The parser still does the same amount of lookaheads as before.

oberblastmeister:
I'm closing this since this pr has been split up. The only thing left to do is error handling.

PR #39, opened 2021-10-02
more ergonomic testing

oberblastmeister:
run `env UPDATE_TESTS=1` to update all `.expect` files. Also automatically creates `.expect` files if they do not exist.


aaronjanse:
Thank you @oberblastmeister!!

I haven't yet tested this locally, but the change itself (making it easier to update tests) looks very helpful.

@oberblastmeister Would you mind removing the non-test-related changes (the formatting and parser changes)?

I'd also be happy to do that for you by adding a partial diff revert commit before merging.

Issue #46, opened 2021-11-08
decimal floats without leading zero causes a parsing failure


efx:
### Describe the bug

decimal floats without leading zero causes a parsing failure
Originally reported upstream in https://github.com/nix-community/nixpkgs-fmt/issues/266. Thanks to @aaronjanse  for initial triage!

Similar to #9. 

### Code Snippet to reproduce

```nix
.5
```

### Expected behavior

I expect this to be parsed as a float regardless of the leading `0`. 

### Additional context

I modified the `dump-ast.rs` to work with stdin.
Then, I ran the following:
```
echo ".5" | cargo run --quiet --example dump-ast
error: unexpected TOKEN_DOT at 0..1, wanted any of [TOKEN_PAREN_OPEN, TOKEN_REC, TOKEN_CURLY_B_OPEN, TOKEN_SQUARE_B_OPEN, TOKEN_DYNAMIC_START, TOKEN_STRING_START, TOKEN_IDENT]
NODE_ROOT 0..3 {
  NODE_APPLY 0..2 {
    NODE_ERROR 0..1 {
      TOKEN_DOT(".") 0..1
    }
    NODE_LITERAL 1..2 {
      TOKEN_INTEGER("5") 1..2
    }
  }
  TOKEN_WHITESPACE("\n") 2..3
}

```

Adding the leading `0.5` works:
```
echo "0.5" | cargo run --quiet --example dump-ast
NODE_ROOT 0..4 {
  NODE_LITERAL 0..3 {
    TOKEN_FLOAT("0.5") 0..3
  }
  TOKEN_WHITESPACE("\n") 3..4
}
```

PR #47, opened 2021-11-08
example for stdin parsing

efx:
When debugging #46 I wanted to narrow down the failing snippet. Using a file was too unwiedly.

You can run this snippet with `echo "0.5" | cargo run --quiet --example from-stdin`

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,13 @@ rnix nowadays uses [@matklad](https://github.com/matklad)'s
 
 Examples can be found in the `examples/` directory.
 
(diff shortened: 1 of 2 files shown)


efx:
CI failure is spurious:
- https://github.com/nix-community/rnix-parser/runs/4141635908?check_suite_focus=true#step:4:4

That failure is from any 2.4 era subcommands being executed without opting into the experimental features. e.g. `nix build`

Ma27:
Mentioning this in the readme would be nice :)

Ma27:
Thanks!

PR #53, opened 2021-11-11
Mark duplicated arguments in argument-patterns as syntax error

### LINKED ISSUE #45: Diversion from nix with duplicate function arguments
### Describe the bug

Nix throws a parse error if the idents in a pattern are duplicates. `rnix-parser` does not throw an error. Seems to be defined [here](https://github.com/NixOS/nix/blob/1e7c796e66a692cd097b155bd0a60fedf20d554f/src/libexpr/parser.y#L145).

### Code Snippet to reproduce

```nix
{a, b, a}: a + b
```

### Expected behavior

Throw a parse error, with the position of the first occurrence of the duplicated identifier.

Ma27:
### Summary & Motivation

To remain consistent with `nix-instantiate --parser`, expressions like

    { a, a }: a

should be marked as invalid.

n/a

### Further context
Closes #45

-----

cc @nerdypepper


efx (approved):
Looks good to my inexperienced `rnix-parser` eyes!
I did pull the branch down and run `cargo test`.

src/parser.rs
```diff
@@ -34,6 +34,8 @@ pub enum ParseError {
     UnexpectedEOF,
     /// UnexpectedEOFWanted is used when specific tokens are expected, but the end of file is reached
     UnexpectedEOFWanted(Box<[SyntaxKind]>),
+    /// UnexpectedDuplicatedArgs is used when formal arguments are duplicated, e.g. `{ a, a }`
+    UnexpectedDuplicatedArgs(TextRange, String),
```
nerdypepper, line 38: Maybe we can just call this just `DuplicateArgs`?

src/parser.rs
```diff
@@ -75,6 +77,15 @@ impl fmt::Display for ParseError {
             ParseError::UnexpectedEOFWanted(kinds) => {
                 write!(f, "unexpected end of file, wanted any of {:?}", kinds)
             }
+            ParseError::UnexpectedDuplicatedArgs(range, ident) => {
+                write!(
+                    f,
+                    "argument `{}' is duplicated in {}..{}",
+                    ident,
+                    usize::from(range.start()),
+                    usize::from(range.end())
```
nerdypepper, line 86: I believe the `Debug` implementation of a `TextRange` produces the same string:
```suggestion
                    "argument `{}` is duplicated in {:?}",
                    ident,
                    range
```

Also, I think that single quote (after the first `{}`) is supposed to be a backtick.

test_data/parser/patterns/8.expect
```diff
@@ -0,0 +1,38 @@
+error: argument `a' is duplicated in 8..9
```
nerdypepper, line 1: I think this might look better as:
```suggestion
error: argument `a` is duplicated in 8..9
```

nerdypepper (commented):
I haven't had the chance to test it out locally, but everything else looks good, thanks for working on this!

src/parser.rs
Ma27: This doesn't compile then.

nerdypepper: Hmm, this does seem to work as expected:
```rust
                write!(
                    f,
                    "argument `{}` is duplicated in {:?}",
                    ident,
                    range
                )
```
Anyhow, the change is not really significant, I think the current state is fine as well.

nerdypepper:
Looks good to me! Thanks again, @Ma27.

PR #57, opened 2021-11-14
refactor(flake): rework using fenix and buildRustPackage

lovesegfault:
- chore(flake): update inputs
- refactor(gitignore): add result, result-*
- feat: add rust-toolchain file
- refactor(flake): rework using fenix and buildRustPackage

### Summary & Motivation

While working on bumping to rowan 0.15 (which requires Rust 2021) I got annoyed
at the current build's lack of pinning to a given Rust release, regardless of
the one pinned in nixpkgs.

With this change we use [fenix](https://github.com/nix-community/fenix) to pin
our Rust to the one in the standard `rust-toolchain.toml` file. This should also
allow us to enforce a MSRV (Minimum Supported Rust Version).

In the process, I also decided to use the rust build tooling already present in
nixpkgs, and then using `inputsFrom` for the `devShell`. Overall, I think this
has led to a cleaner and easier to read flake.

Hopefully none, I suppose if someone were adamant about using an ancient nixpkgs
(before `makeRustPlatform`) things wouldn't work.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -17,11 +17,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1634172192,
-        "narHash": "sha256-FBF4U/T+bMg4sEyT/zkgasvVquGzgdAf4y8uCosKMmo=",
+        "lastModified": 1636800699,
+        "narHash": "sha256-SwbyVxXffu3G2ulJIbTf0iQfqhbGbdml4Dyv5j9BiAI=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "2cf9db0e3d45b9d00f16f2836cb1297bcadc475e",
+        "rev": "2fa862644fc15ecb525eb8cd0a60276f1c340c7c",
         "type": "github"
       },
       "original": {
@@ -40,11 +40,11 @@
     },
     "utils": {
       "locked": {
-        "lastModified": 1610051610,
-        "narHash": "sha256-U9rPz/usA1/Aohhk7Cmc2gBrEEKRzcW4nwPWMPwja4Y=",
+        "lastModified": 1634851050,
+        "narHash": "sha256-N83GlSGPJJdcqhUxSCS/WwW5pksYf3VP1M13cDRTSVA=",
         "owner": "numtide",
         "repo": "flake-utils",
-        "rev": "3982c9903e93927c2164caa727cd3f6a0e6d14cc",
+        "rev": "c91f3de5adaf1de973b797ef7485e441a65b8935",
         "type": "github"
       },
       "original": {
--- a/flake.lock
+++ b/flake.lock
@@ -1,17 +1,23 @@
 {
   "nodes": {
-    "import-cargo": {
+    "fenix": {
+      "inputs": {
+        "nixpkgs": [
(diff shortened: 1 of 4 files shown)


zimbatm (approved):
Sounds good to me. Given that Rust 1.56.1 is in nixpkgs-unstable, it's still viable to use the library in next releases of nixpkgs-fmt.

flake.lock
```diff
@@ -33,18 +39,35 @@
...
     },
+    "rust-analyzer-src": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1636846936,
+        "narHash": "sha256-3xJCEfFdYkdxrQ3MP20ZbcjOBPxuG8m89WxpC4WkeW0=",
+        "owner": "rust-analyzer",
+        "repo": "rust-analyzer",
+        "rev": "766b52b598ad1639e5b815b87dbd3fa3be155219",
+        "type": "github"
```
Ma27, line 55: This seems like quite a big dependency - even though it's only for dev purposes (yes, I live in Germany, we don't have fast Internet here %)). Is there any chance to strip that down a bit?

To describe my issue: I'm using direnv and I have distributed building enabled by default (since it mostly saves time). And after ~25mins I decided to stop waiting for my shell to be ready.

Also regarding what needs to be built: is there any way to cache this?

Ma27 (commented):
First of all, thanks a lot for your work, it's highly appreciated! :heart: 

As posted below, I'm a bit concerned about the size of the dev tools we pull in here now. While I'm not sure if there's any way around the amount of things to download, can we at least introduce some kind of caching (I have to admit I'm not sure if we can force e.g. direnv to use the cachix substituter from `fenix`).

FYI, I'll assign this to the 0.11 milestone I just created since I'd like to publish 0.10 ~nowish as this contains changes to support new syntax features from Nix 2.4.

flake.lock
lovesegfault: IMHO `fenix` should just be caching this to the nix-community cachix instance. I'll ask around how to make that happen.

kamadorueda: There are a few options, one being adding adding the nix-community substituter to the system config, other adding the nix-community url to nixConfig.extra-substituters in the flake.nix, and the last making a `nix --extra-substituters url print-dev-env` in direnv

Ma27:
I'm still preferring to just use what nixpkgs gives us, if we really need it, we can go to nixpkgs-unstable where Rust is always relatively new, so I'm inclined to close this, sorry.

But perhaps I'm wrong, so CCing a few contributors for opinions: @oberblastmeister @nerdypepper @aaronjanse @zseri @darichey

nerdypepper:
My 2 cents; I consume rnix-parser as a library through crates.io or git url, and not through the flake. However, I do use the flake for development of rnix-parser. If the pinning the version of rust is important for the dev shell, I would like these changes, otherwise, I would say we use the rust package from nixpkgs.

Ma27:
I can't remember that we actually needed bleeding-edge features of rustc (also I'm not sure if I'd be too happy about it considering that `rnix-lsp` consumes this library and `rnix-lsp` should be buildable with whatever is currently in nixpkgs), so I'd say we're good with what we have.

PR #58, opened 2021-11-14
chore(cargo): move to 2021 Rust edition

lovesegfault:
### Summary & Motivation

Now that we are using rust >= 1.56 we should move to the new 2021 edition, which
gives us the new resolver and helps shrink our dependency graph a bit.

Users would need rustc >=1.56 to use rnix-parser now, but they don't need to
move to the new edition themselves.

### Further context

Draft since this depends on #57


Ma27:
Depends on #57, so also moving to 0.11.0.

Ma27:
@darichey yes, feel free to do that :)

Ma27:
Taken care of in #91.

PR #67, opened 2022-01-02
Add type for STRING_INTERPOL

### LINKED ISSUE #66: Parsing string after `abort` in `nixpkgs/default.nix` into `ParsedType` fails at `NODE_STRING_INTERPOL`
### Describe the bug

`nixpkgs/default.nix` fails to parse with:
```
text range: 209..227
start line: 6
message: unable to parse node of kind NODE_STRING_INTERPOL
```

### Code Snippet to reproduce

```rust
fn main() {
    use rnix::types::{ParsedType, Wrapper};
    let inp = r#"''

    This version of Nixpkgs requires Nix >= ${requiredVersion}, please upgrade:

    - If you are running NixOS, `nixos-rebuild' can be used to upgrade your system.

    - Alternatively, with Nix > 2.0 `nix upgrade-nix' can be used to imperatively
      upgrade Nix. You may use `nix-env --version' to check which version you have.

    - If you installed Nix using the install script (https://nixos.org/nix/install),
      it is safe to upgrade by running it again:

          curl -L https://nixos.org/nix/install | sh

    For more information, please see the NixOS release notes at
    https://nixos.org/nixos/manual or locally at
    ${toString ./nixos/doc/manual/release-notes}.

    If you need further help, see https://nixos.org/nixos/support.html
  ''"#;
    let parsed = rnix::parse(&inp);
    assert!(parsed.errors().is_empty());
    match ParsedType::try_from(parsed.root().inner().expect("root")) {
        Ok(ParsedType::Str(s)) => {
            for i in s.parts() {
                use rnix::value::StrPart;
                match i {
                    StrPart::Literal(_) => { /* ignore */ }
                    StrPart::Ast(ast) => {
                        let txtrng = ast.text_range();
                        match ParsedType::try_from(ast) {
                            Ok(x) => println!("{:?}", x),
                            Err(e) => println!("ERROR: ({:?}) {}", txtrng, e),
                        }
                    }
                }
            }
        }
        Ok(_) => unreachable!(),
        Err(e) => {
            println!("{:?}", e);
        }
    }
}
```

Output:
```
$ RUST_BACKTRACE=1 cargo run
   Compiling rnixbug v0.1.0 (/home/zseri/tmp/rnixbug)
    Finished dev [unoptimized + debuginfo] target(s) in 0.70s
     Running `/home/zseri/tmp/rnixbug/target/debug/rnixbug`
ERROR: (48..66) invalid cast from kind NODE_STRING_INTERPOL

ERROR: (643..687) invalid cast from kind NODE_STRING_INTERPOL

```

Input: the string after `abort` in `nixpkgs/default.nix`

### Expected behavior
The given Nix code should parse successfully.

### Additional context
`rnix = "0.10.1"`

zseri:
### Summary & Motivation

Fixes #66.

- `rnix::types::StrPart::Ast(SyntaxNode)` -> `rnix::types::StrPart::Ast(StrInterpol)`
- Adds `rnix::types::TypedNode::ParsedType::StrInterpol`

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -12,7 +12,7 @@ use crate::{
 macro_rules! typed {
     ($($kind:expr => $name:ident$(: $trait:ident)*$(: { $($block:tt)* })*),*) => {
         $(
-            #[derive(Clone, Debug)]
+            #[derive(Clone, Debug, PartialEq, Eq, Hash)]
             pub struct $name(SyntaxNode);
 
             impl TypedNode for $name {
@@ -248,6 +248,7 @@ pub enum ParsedType {
     AttrSet(AttrSet),
     KeyValue(KeyValue),
     Str(Str),
+    StrInterpol(StrInterpol),
     UnaryOp(UnaryOp),
     Value(Value),
     With(With),
@@ -270,6 +271,7 @@ impl TryFrom<SyntaxNode> for ParsedType {
             NODE_INHERIT => Ok(ParsedType::Inherit(Inherit::cast(node).unwrap())),
             NODE_INHERIT_FROM => Ok(ParsedType::InheritFrom(InheritFrom::cast(node).unwrap())),
             NODE_STRING => Ok(ParsedType::Str(Str::cast(node).unwrap())),
+            NODE_STRING_INTERPOL => Ok(ParsedType::StrInterpol(StrInterpol::cast(node).unwrap())),
             NODE_LAMBDA => Ok(ParsedType::Lambda(Lambda::cast(node).unwrap())),
             NODE_LEGACY_LET => Ok(ParsedType::LegacyLet(LegacyLet::cast(node).unwrap())),
             NODE_LET_IN => Ok(ParsedType::LetIn(LetIn::cast(node).unwrap())),
@@ -319,6 +321,7 @@ impl TypedNode for ParsedType {
             ParsedType::AttrSet(n) => n.node(),
             ParsedType::KeyValue(n) => n.node(),
             ParsedType::Str(n) => n.node(),
+            ParsedType::StrInterpol(n) => n.node(),
             ParsedType::UnaryOp(n) => n.node(),
             ParsedType::Value(n) => n.node(),
             ParsedType::With(n) => n.node(),
@@ -411,6 +414,7 @@ typed! [
             value::string_parts(self)
         }
     },
+    NODE_STRING_INTERPOL => StrInterpol: Wrapper,
     NODE_LAMBDA => Lambda: {
         /// Return the argument of the lambda
         pub fn arg(&self) -> Option<SyntaxNode> {
--- a/src/value.rs
+++ b/src/value.rs
@@ -332,6 +332,61 @@ mod tests {
         }
     }
     #[test]
+    fn parts_ast() {
+        use crate::types::{ParsedType, Wrapper};
+        use crate::value::StrPart;
+        use std::convert::TryFrom;
+
+        fn assert_eq_ast_ctn(it: &mut dyn Iterator<Item = StrPart>, x: &str) {
+            let tmp = it.next().expect("unexpected EOF");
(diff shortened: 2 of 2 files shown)


zseri:
usage e.g. https://github.com/YZITE/nix2js/commit/9f7bf5786b9321df79e21162ba324275c6a7c3a0

src/value.rs
```diff
@@ -217,10 +216,10 @@ impl Value {
     }
 }
 
-#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+#[derive(Clone, Debug)]
```
Ma27, line 219: It'd be quite nice if we could retain these traits.

src/value.rs
```diff
@@ -321,15 +319,18 @@ mod tests {
...
-                    // Get the below with nix repl
-                    "    \n          \nThis is a multiline string :D\n  indented by two\n\\'\\'\\'\\'\\\n${ interpolation was escaped }\ntwo single quotes: ''\nthree single quotes: '''\n"
-                ))
-            ]
-        );
+                "#
+            .replace("|trailing-whitespace", "")
+            .as_str(),
+        ))[..]
+        {
```
Ma27, line 326: I think that inlining this in an if-condition is somewhat hard to read. Could you split this up into a declaration and a statement, please?

src/value.rs
```diff
@@ -321,15 +319,18 @@ mod tests {
...
+            assert_eq!(lit,
+                // Get the below with nix repl
+                "    \n          \nThis is a multiline string :D\n  indented by two\n\\'\\'\\'\\'\\\n${ interpolation was escaped }\ntwo single quotes: ''\nthree single quotes: '''\n"
```
Ma27, line 329: Given above's change we should also add a test-case for interpolation-nodes.

Ma27 (commented):
Got a few minor comments, but it mostly seems reasonable to me.

Would be interested in what other downstream consumers of this library have to say (cc @nerdypepper @aaronjanse @zimbatm )

src/value.rs
zseri: Does `SyntaxNode` support these traits in a way we would expect?

src/types.rs
```diff
@@ -12,7 +12,7 @@ use crate::{
 macro_rules! typed {
     ($($kind:expr => $name:ident$(: $trait:ident)*$(: { $($block:tt)* })*),*) => {
         $(
-            #[derive(Clone, Debug)]
+            #[derive(Clone, Debug, PartialEq, Eq, Hash)]
```
zseri, line 15: This was necessary to use the derive for `PartialEq, Eq, Hash` for `StrPart`.

src/value.rs
zseri: I agree, but I don't know what exactly we should test.

Ma27: I'd suggest to test the result of a string with an actual interpolation. And perhaps also include something like your code-snippet from #66 in a test :)

src/types.rs
Ma27: LGTM :+1:

nerdypepper:
Understood the bug now, sorry for the late reply. 

I am for this change, none of the breaking changes affect `statix` at the moment.

zimbatm:
:+1: for the more precise annotation.

Since the interface is changing, is there any way to also handle indentation properly? Right now, the indent looks like it's part of the string, and has to be calculated separately.

zseri:
@zimbatm I think indention is somewhat unrelated to this, because `StrPart` afaik (and according to the succeeding tests) already contains the unindented string; so a method to get that information should be added to `Str` instead.

Ma27:
I agree that the indentation is unrelated, but indeed, we should fix that to match upstream.

Going to open an issue for that. Since the next planned release is 0.11.0 anyways (and thus breaking), it seems justifiable to me.

Ma27 (approved):

PR #73, opened 2022-01-27
Experiment: publish crate to NPM using `wasm-pack`

yvan-sraka:
### Summary & Motivation

The main motivation is to be able to use this crate in any JS/TS project, and to package and publish it on [NPM](https://www.npmjs.com/package/rnix-parser). You can find the build artifacts generated into `./pkg` folder [here](https://github.com/yvan-sraka/rnix-parser-pkg)!

:warning: For the need of `wasm-pack` wanted target `wasm32-unknow-unkwnon`, I quickly change the `flake.nix` configuration, this shouldn't be merged as is.

### Further context

We try to build with @RaitoBezarius some experimental GUI frontend for Nix, using [Svelte](https://svelte.dev/) in a first PoC.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -79,6 +79,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "console_error_panic_hook"
+version = "0.1.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a06aeb73f470f66dcdbf7223caeebb85984942f22f1adb2a088cf9668146bbbc"
+dependencies = [
+ "cfg-if",
+ "wasm-bindgen",
+]
+
+[[package]]
 name = "countme"
 version = "2.0.4"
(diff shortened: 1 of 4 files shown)


Ma27:
Please note that I'm not necessarily inclined to maintain such a platform as well. However, if you're still willing to work on this (and help out with maintenance), then I'd be OK with merging it in and advertise the support as "experimental". Please rebase your branch in that case.

zseri:
I'm also not sure if/how this is beneficial (I've wrote [`nix2js`](https://github.com/YZITE/nix2js) stuff), and I have no idea how this should/could be used by downstream JS or WASM code. I think most stuff that actually wants to use it would currently use an approach like the one used in `nix2js` (the transpile function is exposed as an API from a wasm-packed rust module, which internally uses rnix-parser); exposing the rnix parser type hierarchy and functions in a way which is actually usable and performant enough sounds hard and cumbersome.

PR #77, opened 2022-02-19
make UnaryOp::operator return Option<UnaryOpKind>

nerdypepper:
### Summary & Motivation

`UnaryOp::operator` should not panic if the AST is incomplete, and should return `None` instead. This is also now uniform with `BinaryOp::operator`.

Changes the return type of `UnaryOp::operator`.


Ma27 (approved):

Ma27:
@nerdypepper thanks! %)

Issue #81, opened 2022-03-15
Handle `true`, `false` ?


dmadisetti:
it seems to me that they are primitives and should be castable to `rnix::value::Value::Boolean`

aaronjanse:
Note that `true` can be a variable name:
```
nix-repl> let true = 7; in true
7
```

dmadisetti:
Right, I found out in #83

I think this is pretty disappointing. What would thoughts be on removing TOKEN_BOOL, but keeping the conversation? I guess I can implement this out of the library. It's just annoying

darichey:
I don't think there's anything the parser should do here. `true`, `false`, and `null` are essentially variable names which are bound at the top level to their corresponding values, but those names can be shadowed like any other (see [reference impl](https://github.com/NixOS/nix/blob/63df0fda94b60719fd1db13fa42422d0835561ad/src/libexpr/primops.cc#L3893-L3900)). Therefore, the distinction must be made during evaluation.

Ma27:
I agree with @darichey here.

PR #82, opened 2022-03-16
Support for 'true' and 'false' to be interperted as 'Boolean' literals

dmadisetti:
See #81 

### Summary & Motivation

true and false are literal booleans...

true and false are not TOKEN_BOOLEAN opposed to TOKEN_IDENT additionally they are LITERAL kind. Sooo, this may break things? Easy fix, I don't understand why this wasn't already in here...

### DIFF
--- a/src/kinds.rs
+++ b/src/kinds.rs
@@ -55,6 +55,7 @@ pub enum SyntaxKind {
     TOKEN_OR,
 
     // Identifiers and values
+    TOKEN_BOOLEAN,
     TOKEN_DYNAMIC_END,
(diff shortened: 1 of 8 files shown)


dmadisetti:
Note, I did manually update the tests (with sed and some grit) to make sure I didn't break anything else.

samhug:
IIUC, `true` and `false` are parsed as `TOKEN_IDENT` because nix allows things like:
```
nix-repl> let false = "foo"; in false
"foo"
```

dmadisetti:
wow

Ma27:
Wow, TIL %)

Issue #85, opened 2022-05-28
Fix variable-interpolation in paths


Ma27:
### Describe the bug
see https://github.com/nix-community/rnix-parser/pull/52#issuecomment-1140250547

PR #86, opened 2022-05-29
Fix path interpolation

ncfavier:
Fixes https://github.com/nix-community/rnix-parser/issues/85


Ma27:
Changes looking good, thank a lot for taking care of that :heart:  

Will take a closer look at the code and test it a bit, but then we should be good to go with merge & release here :)

Ma27 (approved):

Ma27:
Thanks! Will try to publish a 0.10.2 soonish.

PR #87, opened 2022-06-15
Fix underscore lambda arg parsed as uri

### LINKED ISSUE #76: Parses certain functions as URIs
### Describe the bug

A lambda expr written without any spaces, with an ident as its body is parsed as a URI.

### Code Snippet to reproduce

```nix
a = _:null; # parsed as URI instead of lambda
```

### Expected behavior

Parse as lambda expeession.

darichey:
### Summary & Motivation
Lambdas which used underscores in the argument name with no space after the colon were incorrectly parsed as URIs. This is incorrect, because [the URI scheme cannot contain an underscore](https://github.com/NixOS/nix/blob/master/src/libexpr/lexer.l#L123).

The approach is to collect the "skipped" characters while checking if the current token is a path or URI. In the case of a URI, those skipped chars constitute the URI scheme, so we can check if it contains an underscore.

I added a couple tests to make sure I didn't break handling of URIs or paths with underscores in them.

### Further context
Fixes #76


Ma27 (approved):

Ma27:
Thanks!

PR #88, opened 2022-06-15
Export tokenize function from tokenizer module

darichey:
### Summary & Motivation

The `Tokenizer` doc mentions using the `tokenize` convenience function, but this isn't exported from the module.

This is slightly nicer than constructing and collecting the Tokenizer manually, because the type is explicit. i.e., compare
```rust
let tokens = Tokenizer::new(input).collect::<Vec<(SyntaxKind, SmolStr)>>();
```
vs
```rust
let tokens = tokenize(input);
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,6 +10,7 @@ pub use self::{
     kinds::SyntaxKind,
     parser::AST,
     value::{StrPart, Value as NixValue},
+    tokenizer::tokenize,
 };
 
 pub use rowan::{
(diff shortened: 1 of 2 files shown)


darichey:
We might also consider re-exporting `tokenize` in lib.rs

Ma27:
> We might also consider re-exporting tokenize in lib.rs

Sounds reasonable

darichey:
Done :+1:

PR #89, opened 2022-06-30
refactor: get rid of smol-str dependency

zseri:
### Summary & Motivation
remove smol-str, broken out of #36 (although I've done it from scratch in this PR). => to reduce the scope of #36.

I haven't checked if/where `SmolStr` might've leaked into public API, but I needed to modify a single example and some tests, to this is probably backward-incompatible.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15,9 +15,9 @@ dependencies = [
 
 [[package]]
 name = "autocfg"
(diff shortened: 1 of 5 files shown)


Ma27 (approved):

Ma27:
Thanks!

PR #92, opened 2022-07-01
chore+refactor: upgrade rowan to v0.15

zseri:
### Summary & Motivation

upgrade `rowan` to v0.15, broken out of https://github.com/nix-community/rnix-parser/pull/36 (although I've done it from scratch in this PR). => to reduce the scope of https://github.com/nix-community/rnix-parser/pull/36.

* replaces `TokenWrapper::as_str :: . -> &str` with `TokenWrapper::to_inner_string :: . -> String`. I didn't know how to access the inner green node + children without allocations or dangling references... I added a note about that to the docstring
* `Value::to_value` now uses a different way to retrieve the token kind, which should usually be equivalent, but I don't know if this breaks any edge case

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -184,11 +184,19 @@ pub trait TypedNode: Clone {
 }
 
 pub trait TokenWrapper: TypedNode {
-    fn as_str(&self) -> &str {
-        match &self.node().green().children().next() {
-            Some(rowan::NodeOrToken::Token(token)) => token.text(),
-            _ => unreachable!(),
-        }
+    fn to_inner_token(&self) -> SyntaxToken {
+        self.node()
+            .children_with_tokens()
+            .filter_map(|element| element.into_token())
+            .next()
+            .unwrap()
+    }
+
+    /// ### Migration note about `as_str` (from `v0.10`)
+    /// `TokenWrapper::as_str` usage (`x.as_str()`) should be replaced with (`x.to_inner_string()`).
+    /// Note that the original usage used references, but the new API returns an owned string
+    fn to_inner_string(&self) -> String {
+        self.to_inner_token().text().to_string()
     }
 }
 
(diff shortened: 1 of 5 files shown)


src/types.rs
```diff
@@ -184,11 +184,19 @@ pub trait TypedNode: Clone {
...
+        self.node()
+            .children_with_tokens()
+            .filter_map(|element| element.into_token())
+            .next()
+            .unwrap()
+    }
+
+    /// ### Migration note about `as_str` (from `v0.10`)
+    /// `TokenWrapper::as_str` usage (`x.as_str()`) should be replaced with (`x.to_inner_string()`).
+    /// Note that the original usage used references, but the new API returns an owned string
```
Ma27, line 197: I'm not sure if I understand the problem you're referring to in your op :)

zseri: simple: `to_inner_string` allocates, `as_str` didn't. But `to_inner_string` is still sometimes more convenient than using `.to_inner_token().text()`.

Ma27: * If we say that it should be replaced with something else we de facto deprecate it and should eliminate all references from our code-base.
* Also, can you point me to the dangling-references issue you referenced? To me, this sounds like a problem.

zseri: (a) Most occurrences are just tests/examples. Yeah, we can probably do that. I would move the note to `to_inner_token` then, which would be a "sensible upgrade path" or so...
(b) Not a problem of the code, just the impossibility of pleasing the borrow checker with a direct equivalent to `as_str`. That is, this doesn't introduce any dangling references.

Ma27: >  (b) Not a problem of the code, just the impossibility of pleasing the borrow checker with a direct equivalent to as_str. That is, this doesn't introduce any dangling references.

Ah I see, I guess I misunderstood your initial comment then, sorry!
Going to take a look, but the rest seems good :)

Ma27 (approved):

PR #94, opened 2022-07-10
String references & less code duplication

zseri:
### Summary & Motivation

https://github.com/nix-community/rnix-parser/pull/36#issuecomment-1179791437

I fixed the one location mentioned, another one related, all the places which were affected by that, and then (in the second commit) went off and fixed a big code-duplication in the tokenizer.

afaik none

### Further context

pre-refactor before #36 gets rebased.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -181,8 +181,9 @@ where
     }
     fn drain_trivia_buffer(&mut self) {
         for (t, s) in self.trivia_buffer.drain(..) {
-            self.consumed += TextSize::of(s.as_str());
-            self.builder.token(NixLanguage::kind_to_raw(t), s.as_str())
+            let s = &s;
+            self.consumed += TextSize::of(s);
+            self.builder.token(NixLanguage::kind_to_raw(t), s);
         }
     }
     fn eat_trivia(&mut self) {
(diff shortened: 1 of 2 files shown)


oberblastmeister:
I think `Token` is a better name than `TokenizeItem`

src/parser.rs
```diff
@@ -360,9 +359,9 @@ where
                         let tp = self.get_text_position();
                         let mut ident_done = false;
                         if let Some((_, ident_name)) = self.peek_raw() {
-                            let contains = args.contains_key(ident_name);
-                            let id = ident_name.clone();
```
oberblastmeister, line 364: Use `&str` in the `HashMap` and only clone if there is an error

zseri:
Fixed. As long as no-one objects to the lint/fmt fixes (which arguably don't strictly belong to this PR, but selectively adding cargo fmt fixes turned out to be pretty annoying...), this should be ready to merge.

oberblastmeister:
This looks good to me.

oberblastmeister:
@Ma27 can you merge this so I can rebase #36?

Ma27 (approved):
First of all, thanks to both of you! I guess I underestimated the footprint of String allocations (even though I knew about the properties of SmolStr). I guess we should somehow make the benchmarking part of our CI (including comparisong from PR branches and master).

Regarding your change: looks good and seems way more optimal, so :+1: 

It would be nice if you could refrain from doing linter/formatter changes inside such a PR, it only makes things harder to review. We should probably add a CI step for that as well.

Issue #95, opened 2022-07-11
Can't access all necessary children of PathWithInterpol


darichey:
`PathWithInterpol` is a `Wrapper` in the typed API, but this only provides access to the first child `StrInterpol` node. However, there can be multiple children `TOKEN_PATH`s and `NODE_STRING_INTERPOL`, all of which are necessary to reconstruct the original path.

For example, `cargo run --example from-stdin <<< './a/b/${"c"}/${"d"}/e/f'` produces the following tree:
```
NODE_ROOT 0..24 {
  NODE_PATH_WITH_INTERPOL 0..23 {
    TOKEN_PATH("./a/b/") 0..6
    NODE_STRING_INTERPOL 6..12 {
      TOKEN_INTERPOL_START("${") 6..8
      NODE_STRING 8..11 {
        TOKEN_STRING_START("\"") 8..9
        TOKEN_STRING_CONTENT("c") 9..10
        TOKEN_STRING_END("\"") 10..11
      }
      TOKEN_INTERPOL_END("}") 11..12
    }
    TOKEN_PATH("/") 12..13
    NODE_STRING_INTERPOL 13..19 {
      TOKEN_INTERPOL_START("${") 13..15
      NODE_STRING 15..18 {
        TOKEN_STRING_START("\"") 15..16
        TOKEN_STRING_CONTENT("d") 16..17
        TOKEN_STRING_END("\"") 17..18
      }
      TOKEN_INTERPOL_END("}") 18..19
    }
    TOKEN_PATH("/e/f") 19..23
  }
  TOKEN_WHITESPACE("\n") 23..24
}
```
Currently, it is only possible to access the 'c' `NODE_STRING_INTERPOL` node.

darichey:
Fixed by #96

PR #96, opened 2022-07-11
Implement typed PathPart parsing

darichey:
### Summary & Motivation

A quick attempt at fixing #95.

`PathWithInterpol` is no longer a `Wrapper` node.

### Further context

This might not be exactly the direction we want to go, but it seemed the most straightforward. We might consider combining `StrPart` and `PathPart` considering their overlap (`InterpolationPart` maybe?).


darichey:
It's also worth mentioning that, assuming a parse with no errors, the first part is always going to be a valid path which we could parse into a `Value::Path`. I think it would be nice to expose that, but that would require some bigger refactoring.

darichey:
I would agree, but I wanted to be consistent with `Str::parts`. I'd be happy to change them both.

darichey:
> It's also worth mentioning that, assuming a parse with no errors, the first part is always going to be a valid path which we could parse into a `Value::Path`. I think it would be nice to expose that, but that would require some bigger refactoring.

[`61c0789` (#96)](https://github.com/nix-community/rnix-parser/pull/96/commits/61c0789e9af16d9005d87c172cd78d119b6a386b) implements this with some breaking changes. Again, not sure if this is definitely the right way to go, but just trying things out. I'll wait for some feedback before doing more! :)

zseri:
> I would agree, but I wanted to be consistent with Str::parts. I'd be happy to change them both.

If you're able to do that, please do it (lately, I was working on extracting some changes from @oberblastmeister's more-typed-stuff PR, which involved reducing unnecessary memory allocations, and it would be nice to "keep this spirit", because it might also improve downstream performance, and reduce peak memory usage)

src/ast/path_util.rs
```diff
@@ -0,0 +1,69 @@
...
+        for child in self.syntax().children_with_tokens() {
+            match child {
+                NodeOrToken::Token(token) => {
+                    assert_eq!(token.kind(), TOKEN_PATH);
+                    parts.push(PathPart::Literal(token.text().to_string()));
+                }
+                NodeOrToken::Node(node) => {
+                    assert_eq!(node.kind(), NODE_STRING_INTERPOL);
+                    parts.push(PathPart::Interpolation(
+                        ast::StrInterpol::cast(node.clone()).unwrap(),
```
zseri, line 19: this cast should mean that the assert directly above it is unnecessary, right?

darichey:
Reworked based on @oberblastmeister Path changes in https://github.com/nix-community/rnix-parser/commit/9df0b116c3fcb65607238935896eb660002e3edb. I noticed that commit makes `ast::Path` ambiguous (it could either be the token Path or the node Path). 

I also combined `StrPart` and `PathPart` into `InterpolPart` and return an `impl Iterator` instead of `Vec`.

@zseri I'm going to hold off on changing `Str::parts` in this PR since I believe @Ma27 is making changes there rn.

oberblastmeister:
@darichey You can fix that here. What name should we use?

darichey:
Sorry, I've been away. I see the rename was done in https://github.com/nix-community/rnix-parser/pull/134. Is there anything else that needs to be done here?

oberblastmeister:
@darichey I think `InterpolPart` should be generic over `T`. We should have a new node called `StrContent` or something and `Str::parts` should return `InterpolPart<StrContent>`. I think the existing `Str::parts` methods should be refactored to take an iterator over `InterpolPart<StrContent>` and return a vector of `InterpolPart<String>`, or maybe a tuple.

darichey:
Sorry, I don't understand. Is this what you mean?
```rust
node! { #[from(NODE_STRING_CONTENT)] struct StrContent; }

impl StrContent {
    tg! { content, TOKEN_STRING_CONTENT }
}
```
```rust
enum InterpolPart<T> {
  Literal(T)
  Interpolation(super::Interpol)
}
```
```rust
impl ast::Str {
  pub fn parts(&self) -> impl Iterator<Item = InterpolPart<StrContent>> { /* ... */ }
}
```
```rust
// name and location tbd
pub fn parse_parts(impl Iterator<Item = InterpolPart<StrContent>>) -> impl Iterator<Item = InterpolPart<String>> { /* ... */ }
```
It's not clear to me what is gained here since we have to examine the contents of the inner `TOKEN_STRING_CONTENT` in `ast::Str::parts` anyways.

Also, how should we apply the same ideas to paths?

zseri:
Why do we need the `NODE_STRING_CONTENT` indirection, when it just contains a token, or is this necessary because of some rowan limitation?

oberblastmeister:
`ast::Path::Parts` should return `impl Iterator<Item = InterpolPart<PathContent>>`

oberblastmeister:
@darichey @zseri  `StrContent` should probably be a token

oberblastmeister:
> Sorry, I don't understand. Is this what you mean?

Yes. The idea of returning `InterpolPart<StrContent>` and then converting it to `InterpolPart<String>` is that the former abstracts the manual casting we have currently in `ast::Str::Parts`. In addition, we should have an api for accessing the nodes without further parsing just like we have for other nodes.

darichey:
Got it, I tried implementing that.

> It might be a good idea to expose/return an impl Iterator instead of always allocating a vector, given that this internally already just wraps an iterator in a relatively simple way.

@zseri going way back to this, I just realized that I don't think this is possible for string parts, because we need to iterate over all the children first to calculate the common indent. Do you agree?

src/ast/path_util.rs
```diff
@@ -1,22 +1,33 @@
...
+
+    pub fn parts_parsed(&self) -> Vec<InterpolPart<String>> {
+        self.parts()
+            .map(|part| match part {
+                InterpolPart::Literal(literal) => {
+                    InterpolPart::Literal(literal.syntax().text().to_string())
+                }
+                InterpolPart::Interpolation(interpol) => InterpolPart::Interpolation(interpol),
+            })
+            .collect()
```
zseri, line 29: please omit the `.collect()`, what the user intends with the iterator shouldn't be a business of the library, if avoidable.

oberblastmeister:
@darichey `parts_parsed` should be removed, it doesn't add any value unlike the `Str` variant

oberblastmeister:
I think `normalized_parts` is a better name than `parts_parsed`

zseri:
@darichey
> I just realized that I don't think this is possible for string parts, because we need to iterate over all the children first to calculate the common indent.

yes, correct

PR #99, opened 2022-07-12
Fix various float parsing issues

darichey:
### Summary & Motivation
Fixes parsing of expressions like `.5e1` which are valid floats in Nix.

I simply extracted the logic which applies to similar expressions with leading digits (e.g., `1.5e1`) and called it in both places.

Edit: Now also fixes parsing of expresions like `1.` and `1e+3` (scientific notation with a plus).


ncfavier:
Does this also fix parsing for `1.` ?

darichey:
It does not. I can look into that!

src/tokenizer.rs
```diff
@@ -440,23 +440,27 @@ impl<'a> Tokenizer<'a> {
...
                 }
             }
             _ => TOKEN_ERROR,
         })
     }
+
+    fn consume_scientific(&mut self) -> SyntaxKind {
+        if self.peek() == Some('e') || self.peek() == Some('E') {
+            self.next().unwrap();
+            if self.peek() == Some('-') {
```
oberblastmeister, line 455: This can be `+` also

darichey:
@ncfavier @oberblastmeister both of those issues should be fixed now as well!

ncfavier (approved):

PR #102, opened 2022-07-16
Fix multiline dollar escape

### LINKED ISSUE #100: Multiline dollar escape sequence not handled properly
### Describe the bug
The tokenizer does not correctly handle one of the possible escape sequences for `$` in multiline strings: `''\$`.

### Code Snippet to reproduce
```rust
fn main() {
    for token in rnix::tokenize(r#"''''\${''"#) {
        println!("{token:?}")
    }
}
```

### Expected behavior
```
(TOKEN_STRING_START, "''")
(TOKEN_STRING_CONTENT, "''\\${")
(TOKEN_STRING_END, "''")
```

### Actual behavior
```
(TOKEN_STRING_START, "''")
(TOKEN_STRING_CONTENT, "''\\")
(TOKEN_INTERPOL_START, "${")
(TOKEN_STRING_START, "''")
(TOKEN_ERROR, "")
```

darichey:
### Summary & Motivation
Previously, if the tokenizer encountered the sequence of characters `''\$`, `next_string` would consume the first 3 and then loop back around and see the `$` next, which it treated as the start of an interpolation.

With this change, it will instead always immediately consume the next character after `''\` before looping back. This is correct, because [`''\x` is a valid escape sequence for any `x`](https://github.com/NixOS/nix/blob/59764eb842d0da4f6fcf4ce4b85bf02ac1ae26fc/src/libexpr/lexer.l#L218).

### Further context
Fixes #100

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -124,7 +124,11 @@ impl<'a> Tokenizer<'a> {
                         self.next();
                         self.peek()
                     } {
-                        Some('\'') | Some('\\') | Some('$') => {
+                        Some('\'') | Some('$') => {
+                            self.next().unwrap();
+                        }
+                        Some('\\') => {
+                            self.next().unwrap();
                             self.next().unwrap();
                         }
                         _ => {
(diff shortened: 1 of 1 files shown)


oberblastmeister:
This looks good, thanks for the fix!

PR #103, opened 2022-07-19
Correctly remove multiline trailing whitespace

darichey:
### Summary & Motivation
Previously, trailing whitespace on the last line of a multiline string was always removed, but this is not necessarily correct. For example, the string `''hello    ''` should keep its trailing whitespace.

This PR attempts to mimic the [logic used in the reference implementation of the parser](https://github.com/NixOS/nix/blob/fbd0a6c6e2e87f6679fe5cabaddaa877cf3e5a90/src/libexpr/parser.y#L268-L274) to instead only remove the final line if it consists solely of whitespace.

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -142,17 +142,6 @@ pub fn remove_indent(input: &str, initial: bool, indent: usize) -> String {
     }
     output
 }
-/// Remove any trailing whitespace from a string
-pub fn remove_trailing(string: &mut String) {
-    let trailing: usize = string
-        .chars()
-        .rev()
-        .take_while(|&c| c != '\n' && c.is_whitespace())
-        .map(char::len_utf8)
-        .sum();
-    let len = string.len();
-    string.truncate(len - trailing);
-}
 
 /// An error that occured when parsing a value from a string
 #[derive(Clone, Debug, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


src/value.rs
```diff
@@ -272,9 +261,14 @@ pub(crate) fn string_parts(string: &types::Str) -> Vec<StrPart> {
         if let StrPart::Literal(ref mut text) = part {
             if multiline {
                 *text = remove_indent(text, i == 0, common);
+                // If the last literal is of the form `X\nY`, exclude Y if it consists solely of spaces
                 if i == literals - 1 {
-                    // Last index
-                    remove_trailing(text);
+                    match text.rfind('\n') {
+                        Some(p) if text[p + 1..].chars().all(|c| c == ' ') => {
```
zseri, line 267: This does not recognize tabs and such as whitespace, I assume this is intentional.

darichey: Right, really I should say "spaces" and not "whitespace." If the last line has some other whitespace, like a tab character, it should not be removed. This matches the behavior of the reference implementation:
```
# test.nix
''
hello
	'' # this is a tab
```
```
$ nix-instantiate --parse test.nix
"hello\n\t"
```

Ma27:
If we do this, we should do it properly from the start, i.e. remove all whitespace prefixes. In fact I have started on a patch for that, but it needs some more polishing, expect a PR ~next week :)

zseri:
don't we already do that? The only difference necessary is that the first and last line are handled differently (in them, if they contain only spaces `' '`, they're omitted)

darichey:
Right, I believe that is already handled. That said, I also think that this entire logic could use a rework. I started by trying to just dumbly port the reference implementation in this commit: https://github.com/darichey/rnix-parser/commit/d56b15ff4ae9e1c03a28c1f0894f6881761925dd

I confirmed that it at least parses all multiline strings in nixpkgs (see #107), but honestly the code is kind of ugly, and I haven't done any kind of benchmarks, so I didn't bother making a PR yet. Please feel free to use it as inspiration @Ma27 :)

PR #104, opened 2022-07-19
Derive Debug and Eq for AST

oxalica:
### Summary & Motivation

Derive more auto trait to make life easier.

No.

### Further context

`Eq` is required by auto-memorizing libraries to trace if the AST is changed.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -19,7 +19,7 @@ use crate::{
 const OR: &str = "or";
 
 /// An error that occurred during parsing
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


src/parser.rs
```diff
@@ -93,7 +93,7 @@ impl fmt::Display for ParseError {
 impl std::error::Error for ParseError {}
 
 /// The result of a parse
-#[derive(Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
zseri, line 96:
```suggestion
#[derive(Clone, Debug, PartialEq, Eq)]
```

oxalica:
After API changes in #105, now AST wrappers already implements `Debug, Eq`. :tada:

PR #105, opened 2022-07-19
Better typed api

oberblastmeister:
Continuation of #36


src/ast.rs
```diff
@@ -0,0 +1,92 @@
...
+
+pub trait AstNode: rowan::ast::AstNode<Language = NixLanguage> {}
+
+impl<T> AstNode for T where T: rowan::ast::AstNode<Language = NixLanguage> {}
+
+#[derive(Debug, Clone)]
+pub struct AstNodeChildren<N> {
+    inner: SyntaxNodeChildren,
+    _p: PhantomData<N>,
+}
```
oxalica, line 30: This is [`rowan::ast::AstChildren`](https://docs.rs/rowan/0.15.6/rowan/ast/struct.AstChildren.html), which is returned by [`rowan::ast::support::children`](https://docs.rs/rowan/0.15.6/rowan/ast/support/fn.children.html)

src/ast/nodes.rs
```diff
@@ -0,0 +1,443 @@
...
+    } => Expr,
+
+    NODE_LITERAL => Literal,
+
+    {
+        NODE_DYNAMIC => Dynamic,
+        NODE_STRING => Str,
+    } => Key,
+
+     NODE_IDENT => Ident: {
```
oxalica, line 165: There are quite some indentation errors inside this macro. Please reformat them.
```suggestion
    NODE_IDENT => Ident: {
```

src/ast/nodes.rs
```diff
@@ -0,0 +1,443 @@
...
+        pub fn lambda(&self) -> Option<Expr> {
+            first(self)
+        }
+        /// Return the value which the lambda is being applied with
+        pub fn argument(&self) -> Option<Expr> {
+            nth(self, 1)
+        }
+    },
+     NODE_ASSERT => Assert: {
+        pub fn assert(&self) -> Option<SyntaxToken> {
```
oxalica, line 182: To be consistent,
```suggestion
        pub fn assert_token(&self) -> Option<SyntaxToken> {
```

src/ast/nodes.rs
```diff
@@ -0,0 +1,443 @@
...
+        pub fn body(&self) -> Option<Expr> {
+            nth(self, 1)
+        }
+    },
+    //  NODE_KEY => Key: {
+        // /// Return the path as an iterator of identifiers
+        // pub fn path<'a>(&'a self) -> impl Iterator<Item = SyntaxNode> + 'a {
+        //     self.node().children()
+        // }
+    // },
```
oxalica, line 201: Isn't it already defined in line 163?

src/ast/nodes.rs
```diff
@@ -0,0 +1,443 @@
...
+        pub fn has_ellipsis(&self) -> bool {
+            self.ellipsis_token().is_some()
+        }
+
+        /// Returns the identifier bound with {}@identifier if it exists
+        pub fn pat_bind(&self) -> Option<PatBind> {
+            first(self)
+        }
+    },
+     NODE_ROOT => Root,
```
oxalica, line 390:
```suggestion
    NODE_ROOT => Root: {
        fn expr(&self) -> Option<Expr> {
            first(self)
        }
    }
```

src/macros.rs
```diff
@@ -10,6 +10,7 @@ macro_rules! T {
     (rec)     => ($crate::SyntaxKind::TOKEN_REC);
     (then)    => ($crate::SyntaxKind::TOKEN_THEN);
     (with)    => ($crate::SyntaxKind::TOKEN_WITH);
+    (ident)   => ($crate::SyntaxKind::TOKEN_IDENT);
```
oxalica, line 13: `ident` is not a literal keyword. This is confusing.
Just using `TOKEN_IDENT` when needed should be fine.

```suggestion
```

src/ast/expr_ext.rs
```diff
@@ -0,0 +1,42 @@
+use crate::{
+    ast::{self, support::*, AstNode, AstToken},
+    kinds::SyntaxKind::*,
+    match_ast, SyntaxElement,
+};
+
+use super::support::first;
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub enum LiteralKind {
```
oxalica, line 13: Could we simply define `enum Literal` like `enum Expr` and remove `NODE_LITERAL`?

src/ast/nodes.rs
```diff
@@ -0,0 +1,449 @@
...
+        //     self.node().children()
+        //         .find_map(InheritFrom::cast)
+        // }
+        // /// Return all the identifiers being inherited
+        // pub fn idents(&self) -> impl Iterator<Item = Ident> {
+        //     self.node().children().filter_map(Ident::cast)
+        // }
+    },
+
+     NODE_INHERIT_FROM => InheritFrom: { },
```
oxalica, line 263:
```suggestion
     NODE_INHERIT_FROM => InheritFrom: {
         fn l_curly_token(&self) -> Option<SyntaxToken> { ... }
         fn expr(&self) -> Option<Expr> { ... }
         fn r_curly_token(&self) -> Option<SyntaxToken> { ... }
     },
```

src/ast/nodes.rs
```diff
@@ -0,0 +1,449 @@
...
+            first(self)
+        }
+
+        /// Return the body of the lambda
+        pub fn body(&self) -> Option<Expr> {
+            first(self)
+        }
+    },
+     NODE_LEGACY_LET => LegacyLet: EntryHolder,
+     NODE_LET_IN => LetIn: EntryHolder: {
```
oxalica, line 282: `fn {let,in}_token`

src/ast/nodes.rs
```diff
@@ -0,0 +1,449 @@
...
+     },
+
+     NODE_PAT_BIND => PatBind: {
+        /// Return the identifier the set is being bound as
+        pub fn ident(&self) -> Option<Ident> {
+            first(self)
+        }
+    },
+
+     NODE_PAT_ENTRY => PatEntry: {
```
oxalica, line 350: `fn question_token`

src/ast/nodes.rs
```diff
@@ -0,0 +1,449 @@
...
+        /// Return the identifier the argument is being bound as
+        pub fn ident(&self) -> Option<Ident> {
+            first(self)
+        }
+
+        /// Return the default value, if any
+        pub fn default(&self) -> Option<SyntaxNode> {
+            self.syntax().children().nth(1)
```
oxalica, line 358: Isn't it `-> Option<Expr>`?

oxalica (changes requested):

src/macros.rs
oberblastmeister: It doesn't matter that `ident` is not a literal keyword, as the macro just gets tokens. I don't think it is that confusing. A note can be put above the macro if needed

oberblastmeister:
That was a lot of work but I think I finished.

src/ast/expr_ext.rs
oberblastmeister: There is now a comment that explains

examples/list-fns.rs
```diff
@@ -3,70 +3,72 @@ use std::{env, error::Error, fs};
...
-        None => {
-            eprintln!("Usage: list-fns <file>");
-            return Ok(());
-        }
-    };
-    let content = fs::read_to_string(&file)?;
-    let ast = rnix::parse(&content).as_result()?;
-    let set = ast.root().inner().and_then(AttrSet::cast).ok_or("root isn't a set")?;
+    todo!()
+//     let file = match env::args().nth(1) {
```
zseri, line 7: instead of commenting every line separately, it would be more appropriate to just use comment blocks.

src/ast.rs
```diff
@@ -0,0 +1,72 @@
...
+
+impl<T> AstNode for T where T: rowan::ast::AstNode<Language = NixLanguage> {}
+
+mod support {
+    use rowan::ast::AstChildren;
+
+    use super::{AstNode, AstToken};
+    use crate::{SyntaxElement, SyntaxKind, SyntaxNode, SyntaxToken};
+
+    pub(super) fn first<N: AstNode, NN: AstNode>(parent: &N) -> Option<NN> {
```
zseri, line 30: I think just using `pub` instead of `pub(super)` would look better. `mod support` isn't exported anywhere anyways, so it would make a difference anyways.

src/ast/expr_ext.rs
```diff
@@ -0,0 +1,45 @@
...
+            return LiteralKind::Integer(it);
+        }
+
+        if let Some(it) = token(self) {
+            return LiteralKind::Path(it);
+        }
+
+        if let Some(it) = token(self) {
+            return LiteralKind::Uri(it);
+        }
```
zseri, line 41: this code doesn't make sense yet.

oberblastmeister: Type inference is one of the best things about rust. It is common to just `.into` without using annotations. What is confusing to you?

oberblastmeister: Type inference is one of the best things about rust. It is common to just `.into` without using annotations. What is confusing to you?

zseri: The problem is that this looks like weird repetition, and it is necessary to know that `token` is generic over the return parameter to be able to understand what's going on. Transforming it into an `else if` chain would also make the purpose a bit clearer, maybe also add a comment.

e.g.
```rust
        if let Some(it) = token(self) {
            return LiteralKind::Float(it);
        } else if let Some(it) = token(self) {
            return LiteralKind::Integer(it);
        } else if let Some(it) = token(self) {
            return LiteralKind::Path(it);
        } else if let Some(it) = token(self) {
            return LiteralKind::Uri(it);
        }
```

could we instead match on the token type directly?

examples/list-fns.rs
oberblastmeister: I'm not doing examples yet

src/ast/expr_ext.rs
oberblastmeister: I don't see how the else if makes it better. If it is confusing, you can just hover over the type. `rust-analyzer` also shows really nice inlay type annotations, which clearly show the different types.

zseri: My complaint is that most people who try to look into the code will stumble over this, especially when unfamiliar, at least it makes the code harder to review.

oberblastmeister: I will accept a comment

zseri:
Tests should be ported before this should/can be merged.

oberblastmeister:
The tests should be ported already, hence the massive diff

oberblastmeister:
Also `rowan` already has a pretty printer for the syntax tree, so we can remove our `TextDump` type. I was going to update the tests with the `rowan` pretty printer but the diff got too large. I will probably do that in another pr.

oxalica:
> I'm going to remove Value, as it is replaced by LiteralKind. The library is focused on syntax and not semantics, so Value is not appropriate here. 

+1. I'd say `mod value` should be simply thrown away since the semantic-relative things are out of scope of this crate. If we do want to keep unescaping stuff, they should be auxiliary methods under `struct ast::Str`.

Also `mod types` would be superseded (and probably removed) by `mod ast` introduced in this PR.

darichey:
I definitely think the unescaping, dedenting, and interpolation stuff should remain. imo they are parsing tasks. Also, the logic for those is non-trivial and would have to be reimplemented by any user interested in string contents.

oberblastmeister:
> I definitely think the unescaping, dedenting, and interpolation stuff should remain. imo they are parsing tasks. Also, the logic for those is non-trivial and would have to be reimplemented by any user interested in string contents.

They are just moved to methods on `ast::Str`

oberblastmeister:
@oxalica the issue with using the rowan `AstNode` trait is that you will have to write `Language = NixLanguage` in every bound, which is undesirable. I solve this by doing a blanket impl with our own `AstNode`. But now to use `AstNode` you have to still import the rowan one. I feel like we should just go with our own trait. It's not much extra code.

oxalica:
@oberblastmeister
> the issue with using the rowan `AstNode` trait is that you will have to write `Language = NixLanguage` in every bound, which is undesirable.

That's desirable.
The advantage of using upstream trait is that downstream users don't have to learn and import similar traits from different crates. Upstream trait provides a consistent interface and can also make generics (or interop) between downstream crates easier.

> I solve this by doing a blanket impl with our own `AstNode`. But now to use `AstNode` you have to still import the rowan one. I feel like we should just go with our own trait. It's not much extra code.

I prefer no-reinventing the wheel than extra codes. The majority of extra codes can be and is already solved by macros. We *could* do a trait alias (via supertrait + blanket impl you did), but users only using upstream `rowan::AstNode::cast` or friends, should still work fine.

src/ast/expr_ext.rs
oxalica: :thinking: I don't consider `Str` is a *Literal* in my own parser crate. Because the interpolation making it not *literally* mean the text of it.
Well, that's arguable.

darichey: What if we had a dedicated node for each of these which is otherwise just a token? e.g., NODE_INTEGER which just has a TOKEN_INTEGER child. Then we don't need to define these manually, and we don't need to answer the question of which of them are "literals".

src/lib.rs
```diff
@@ -38,72 +34,136 @@ impl rowan::Language for NixLanguage {
 pub type SyntaxNode = rowan::SyntaxNode<NixLanguage>;
 pub type SyntaxToken = rowan::SyntaxToken<NixLanguage>;
 pub type SyntaxElement = rowan::NodeOrToken<SyntaxNode, SyntaxToken>;
+pub type SyntaxElementChildren = rowan::SyntaxElementChildren<NixLanguage>;
+pub type SyntaxNodeChildren = rowan::SyntaxNodeChildren<NixLanguage>;
 
-/// A convenience function for first tokenizing and then parsing given input
-pub fn parse(input: &str) -> AST {
-    parser::parse(Tokenizer::new(input))
-}
```
darichey, line 45: I think this should say, even if it's just an alias for `Root::parse`. It provides a more obvious "entry point" to the crate.

oberblastmeister:
I'm thinking of deleting `Root`, as it provides no value. It's just a newtype over `Expr`.

oxalica:
@oberblastmeister 
> I'm thinking of deleting `Root`, as it provides no value. It's just a newtype over `Expr`.

No. `Root` wraps white-spaces outside the top-level `Expr`, which is required for lossless parsing.

https://github.com/nix-community/rnix-parser/blob/f759fce8ac005ed14d13219579e62c06bdfba678/test_data/general/docs.expect#L1-L4

oberblastmeister:
Woops completely forgot. It's been too long since I did parsing with rowan.

src/lib.rs
oberblastmeister: We can just document that `Root::parse` is the entry point.

oberblastmeister:
> I prefer making low-level node structure as simple and flat as possible. Thus, NODE_LITERAL only wraps immediate tokens (to make up Expr::Literal, not Expr::{Int,Float}, for convenience)
NODE_STR directly makes up Expr::Str as a non-terminating syntax structure. Not under enum Literal.
Also this is what we've done in master. So we can avoid updating test data in this PR.

@oxalica
Here's what I think we should do. Remove the `Literal` type and `NODE_LITERAL` variant altogether, and rename `LiteralKind` to `Literal`.

oberblastmeister:
Actually that might not work with the macro, I'll have to think more about this. I definitely agree with you about making the nodes as simple and flat as possible.

src/ast/expr_ext.rs
oberblastmeister: What if we get rid of the concept of a `Literal` and inline it into `Expr`. This would be the most simple. I don't feel like the literal concept is very useful seeing as different things can be interpreted as a literal. I think this is similar to how `Value` was not very useful.

oberblastmeister:
Forgot that I could click *show hidden conversations*

src/ast/expr_ext.rs
oxalica: > What if we get rid of the concept of a Literal and inline it into Expr.

I'd stick on current node structure and not touching the parser in this PR, whose title is "Better typed api". You can definitely open a later PR for it after this one getting merged.

darichey: Good point, we should stay on track in this PR and decide this later.

examples/list-fns.rs
```diff
@@ -1,12 +1,12 @@
 use std::{env, error::Error, fs};
 
+use rnix::ast::AstToken;
 use rnix::{
     ast::{self, EntryHolder},
```
zseri, line 5:
```suggestion
    ast::{self, AstToken, EntryHolder},
```

zseri (changes requested):

examples/list-fns.rs
```diff
@@ -1,6 +1,20 @@
 use std::{env, error::Error, fs};
 
-use rnix::{types::*, NodeOrToken, SyntaxKind::*, SyntaxNode};
+use rnix::ast::AstToken;
```
zseri, line 3:
```suggestion
```

src/ast.rs
```diff
@@ -0,0 +1,76 @@
...
+
+impl From<Another> for Option<My> {
+    fn from(_: Another) -> Self {
+        None
+    }
+}
+
+fn test() {
+    let test = <Option<My>>::from(Another).unwrap();
+}
```
zseri, line 76: If this is supposed to be a unit-test (and it looks like it should), it should have a `#[test]` attribute (although it would, in it's current state, always panic, so I assume `#[test(should_panic)]`would be more appropriate)

src/ast/operators.rs
```diff
@@ -0,0 +1,68 @@
...
+            TOKEN_LESS => Some(BinOpKind::Less),
+            TOKEN_LESS_OR_EQ => Some(BinOpKind::LessOrEq),
+            TOKEN_MORE => Some(BinOpKind::More),
+            TOKEN_MORE_OR_EQ => Some(BinOpKind::MoreOrEq),
+            TOKEN_NOT_EQUAL => Some(BinOpKind::NotEqual),
+            TOKEN_OR => Some(BinOpKind::Or),
+
+            _ => None,
+        }
+    }
```
zseri, line 50: imo this should be a `core::convert::TryFrom` impl

src/ast/operators.rs
```diff
@@ -0,0 +1,68 @@
...
+
+impl UnaryOpKind {
+    /// Get the operation kind from a token in the AST
+    pub fn from_kind(kind: SyntaxKind) -> Option<Self> {
+        match kind {
+            TOKEN_INVERT => Some(UnaryOpKind::Invert),
+            TOKEN_SUB => Some(UnaryOpKind::Negate),
+            _ => None,
+        }
+    }
```
zseri, line 67: this should also be a `TryFrom` impl

src/ast/str_util.rs
```diff
@@ -0,0 +1,306 @@
...
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub enum StrPart {
+    Literal(String),
+    Interpolation(ast::StrInterpol),
+}
+
+/// Interpret escape sequences in the nix string and return the converted value
+pub fn unescape(input: &str, multiline: bool) -> String {
+    let mut output = String::new();
```
zseri, line 82:
```suggestion
    let mut output = String::with_capacity(input.len());
```

src/ast/str_util.rs
```diff
@@ -0,0 +1,306 @@
...
+    let mut input = input.chars().peekable();
+    loop {
+        match input.next() {
+            None => break,
```
zseri, line 86:
```suggestion
    while let Some(i) = input.next() {
        match i {
```
and remove `Some` from the other branches at this level.

src/ast/str_util.rs
```diff
@@ -0,0 +1,306 @@
...
+
+    remove_indent(input, true, common)
+}
+/// Remove a specified max value of indention from each line in a string after
+/// a specified starting point
+pub fn remove_indent(input: &str, initial: bool, indent: usize) -> String {
+    let mut output = String::new();
+    let mut start = 0;
+    if initial {
+        // If the first line is whitespace, ignore it completely
```
zseri, line 153: afaik this should also apply to the last line

src/ast/str_util.rs
```diff
@@ -0,0 +1,306 @@
...
+        let end = input[start..].find('\n').map(|i| start + i + 1);
+        {
+            let end = end.unwrap_or(input.len());
+            output.push_str(&input[start..end]);
+        }
+        start = match end {
+            Some(end) => end,
+            None => break,
+        };
+    }
```
zseri, line 177: It might be a good idea to rewrite this part to leverage `input.lines()` instead.
e.g.
```rust
    input.lines()
        .flat_map(|i| {
            let expanded_indent = indention(i).take(indent).map(char::len_utf8).sum::<usize>();
            &i[expanded_indent..]
        })
        .collect()
```
Also, could we put `it.map(char::len_utf8).sum::<usize>()` into a separate function?

src/ast/str_util.rs
```diff
@@ -0,0 +1,306 @@
...
+pub fn remove_trailing(string: &mut String) {
+    let trailing: usize = string
+        .chars()
+        .rev()
+        .take_while(|&c| c != '\n' && c.is_whitespace())
+        .map(char::len_utf8)
+        .sum();
+    let len = string.len();
+    string.truncate(len - trailing);
+}
```
zseri, line 191: As noted in another PR, this code actually does prune too much white space (#103)

src/ast/str_util.rs
```diff
@@ -0,0 +1,306 @@
...
+                // Get the below with nix repl
+                "    \n          \nThis is a multiline string :D\n  indented by two\n\\'\\'\\'\\'\\\n${ interpolation was escaped }\ntwo single quotes: ''\nthree single quotes: '''\n"
+            );
+        } else {
+            unreachable!();
+        }
+    }
+
+    //     #[test]
+    //     fn parts_ast() {
```
zseri, line 247: This test checked for unintentional truncation (which happens when the string unescaping parser breaks out too early). This intend should be covered by a test to prevent it from re-occuring (see also #66)

src/ast/nodes.rs
```diff
@@ -0,0 +1,394 @@
...
+            }
+
+            fn cast(from: SyntaxNode) -> Option<Self> {
+                if Self::can_cast(from.kind()) {
+                    Some(Self(from))
+                } else {
+                    None
+                }
+            }
+            fn syntax(&self) -> &SyntaxNode {
```
zseri, line 60:
```suggestion

            #[inline(always)]
            fn syntax(&self) -> &SyntaxNode {
```

src/ast/tokens.rs
```diff
@@ -0,0 +1,73 @@
...
+
+            fn cast(from: SyntaxToken) -> Option<Self> {
+                if from.kind() == $kind {
+                    Some(Self(from))
+                } else {
+                    None
+                }
+            }
+
+            fn syntax(&self) -> &SyntaxToken {
```
zseri, line 38:
```suggestion
            #[inline(always)]
            fn syntax(&self) -> &SyntaxToken {
```

zseri (changes requested):

darichey:
I think any problems with `str_util.rs` should be ignored in this PR, since it's just moved from the now-removed `value.rs`. Those can be a focus of a future PR.

zseri:
ok

examples/list-fns.rs
```diff
@@ -41,32 +74,23 @@ fn main() -> Result<(), Box<dyn Error>> {
...
-        .iter()
-        .map(|it| it.trim_start_matches('#').trim())
+            _ => None,
+        })
+        .flatten()
+        .map(|s| s.text().trim().to_string())
         .collect::<Vec<_>>()
-        .join("\n        ");
-    Some(doc).filter(|it| !it.is_empty())
+        .join(" ")
```
zseri, line 95: do we really want to join them with a space and not newline?

src/lib.rs
```diff
@@ -38,72 +34,136 @@ impl rowan::Language for NixLanguage {
...
+    }) => {{
+        $( if let Some($it) = ast::$ast::cast($node.clone()) { $res } else )*
+        { $catch_all }
+    }};
+}
+
+#[cfg(test)]
+mod tests {
+    // #[test]
+    // fn interpolation() {
```
zseri, line 112: This test checks that interpolation and dedention work correctly when interleaved. As it needs to check the resulting AST (because errors in the past happened in the AST-forming process), it should be ported.

oberblastmeister:
I wonder if we should be using `TryFrom`, our own trait, or just normal methods. All of our uses of `TryFrom` basically have `type Error = ()` which makes it isomorphic to Option.

src/ast/tokens.rs
oberblastmeister: I don't think inlining here is appropriate, and definately not `#[inline(always)]`.

zseri:
We could also try `From<ExtType> for Option<OurType>`, but it might not work due to orphan rules.

oberblastmeister:
I think this is basically ready. This pr is really big and I don't want it to drag on forever. It doesn't have to be perfect, as we can make followup prs. I think currently it is good enough. I want to get this in so I can merge other prs. So @oxalica @darichey @zseri final round of comments before I merge?

src/ast.rs
```diff
@@ -0,0 +1,76 @@
...
+        children_tokens(parent).find(|it| it.kind() == kind)
+    }
+
+    pub(super) fn children_tokens<N: AstNode>(parent: &N) -> impl Iterator<Item = SyntaxToken> {
+        parent.syntax().children_with_tokens().filter_map(SyntaxElement::into_token)
+    }
+
+    pub(super) fn children_tokens_u<N: AstNode>(parent: &N) -> impl Iterator<Item = SyntaxToken> {
+        parent.syntax().children_with_tokens().filter_map(SyntaxElement::into_token)
+    }
```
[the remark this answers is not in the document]
zseri: I guess this is fixed now.

src/ast/operators.rs
zseri: postponed.

src/ast/operators.rs
zseri: postponed.

examples/list-fns.rs
zseri: Fixed.

Ma27:
Would you mind summarizing the changes in here, especially the breaking things? I actually wanted to review it as well, however I didn't have the time for it last week, next week perhaps.

The thing is, this may be fine implementation-wise, but to me it seems as if this change was punched through way too fast. Talking about the API changes etc. first might've been preferable.

oberblastmeister:
I understand your concern, but we talked about the api a lot before merging this, hence the massive comment length. This pr was here for more than a year, and I didn't want to wait any longer, so I don't think it was too fast.

### Summary

- removes the `types` module and replaces it with `ast`
  - the `ast` modules has better types and sum types
  - biggest change is removes `ParsedType` in favor of `Expr`
- removes the `value` module. methods on `Value` go on the individual types. For example, string related methods go on `ast::Str`

asymmetric:
@oberblastmeister after this change, how do I get the root node of an AST, i..e [this](https://github.com/nix-community/nixpkgs-fmt/blob/master/src/lib.rs#L96)?

PR #106, opened 2022-07-19
Rework and fix Attrpath and Dynamic parsing

### LINKED ISSUE #23: Doesn't parse valid Nix edge-cases
`let b = a: a; "or" = 5; in  [b or]` fails to be properly parsed, and instead generates a partial `NODE_OR_DEFAULT` and errors.

This seems to also influence other bits, e.g. `[b or b]` parses as a valid `NODE_OR_DEFAULT`, while it is actually parsed as `[(b or) b]` by Nix.

oxalica:
### Summary & Motivation
Rework and fix Attrpath and Dynamic parsing

Current handling of Attrpath and Dynamic `${` is mostly incorrect.

- The dynamic attribute `${` can only appear as an Attr inside an Attrpath,
  and itself cannot make up an expression.

  But currently we parse it as a kind of Expr.

  This commit reject this usage.

- Dot separated attributes, aka. Attrpath or Key, can appear in 3 places:

  1. After a "." in a Select expression
     `<expression>.<attrpath>`

     https://github.com/NixOS/nix/blob/2.10.3/src/libexpr/parser.y#L439

  2. After a "?" in a HasAttr expression
     `<expression> ? <attrpath>`

     https://github.com/NixOS/nix/blob/2.10.3/src/libexpr/parser.y#L416

  3. As a binding key in Let or Attrset-like expression
     `<attrpath> = <expression>;`

     https://github.com/NixOS/nix/blob/2.10.3/src/libexpr/parser.y#L541

  To make it consistent, all usage above are now wrapped in a
  NODE_ATTRPATH,
  which consists of multiple {NODE_DYNAMIC,NODE_IDENT,NODE_STRING}
  interspersed by TOKEN_DOT.

  NODE_KEY_VALUE is also renamed to NODE_ATTRPATH_VALUE.

  Benefiting from flat Attrpath instead of nested one, we can now get
  rid of NODE_OR_DEFAULT, since it can be trivially distinguished by
  checking if there is an `or` in NODE_ATTRPATH_VALUE.

- `NODE_OR_DEFAULT` is removed
- `NODE_KEY` -> `NODE_ATTRPATH`, `NODE_KEY_VALUE` -> `NODE_ATTRPATH_VALUE`
- `BinOp::IsSet` is removed since it's actually not a normal binary operator.
- `NODE_SELECT` is flatten in favor of `NODE_ATTRPATH` which consists of multiple Attrs, including `a.b.c` and `a.b.c or expr`
- `a or b` is Apply, not "OrDefault", which matches the result of Nix. Fixes #23
- `${` is considered invalid at Expr places now, which matches the result of Nix.

### Further context

~~The name of `NODE_KEY` is weird to me. It's called `attrpath` [in Nix](https://github.com/NixOS/nix/blob/2.10.3/src/libexpr/parser.y#L580) and [in nixpkgs](https://github.com/NixOS/nixpkgs/search?q=attrpath).~~
~~Should we rename it? This would also cause API breakage.~~

Renamed.


darichey (changes requested):
Awesome! This has been on my todo list for a bit :)

> The name of NODE_KEY is weird to me. It's called attrpath [in Nix](https://github.com/NixOS/nix/blob/2.10.3/src/libexpr/parser.y?rgh-link-date=2022-07-19T15%3A19%3A08Z#L580) and [in nixpkgs](https://github.com/NixOS/nixpkgs/search?q=attrpath).
> Should we rename it?

+1 to renaming

> a or b is Apply, not "OrDefault", which matches the result of Nix.

Does this fix #23?

oxalica:
> Does this fix https://github.com/nix-community/rnix-parser/issues/23?

Yes. I referenced it in the description now.

oxalica:
> > The name of NODE_KEY is weird to me. It's called attrpath [in Nix](https://github.com/NixOS/nix/blob/2.10.3/src/libexpr/parser.y?rgh-link-date=2022-07-19T15%3A19%3A08Z#L580) and [in nixpkgs](https://github.com/NixOS/nixpkgs/search?q=attrpath).
> > Should we rename it?
> 
> +1 to renaming

I'd prefer to keep this PR focused on parsing fixes, and leave the renaming to another PR later. Maybe after the refactoring of high-level AST in #105.

darichey (changes requested):
Last thing from me: The`impl TryFrom<SyntaxNode> for ParsedType` should be updated to account for `ParsedType::HasAttr` as well.

src/parser.rs
```diff
@@ -696,11 +693,18 @@ where
         }
         checkpoint
     }
-    fn parse_isset(&mut self) -> Checkpoint {
-        self.handle_operation_left(false, Self::parse_negate, &[TOKEN_QUESTION])
+    fn parse_hasattr(&mut self) -> Checkpoint {
+        let checkpoint = self.parse_negate();
+        while self.peek().map(|t| t == TOKEN_QUESTION).unwrap_or(false) {
```
zseri, line 698: how would this change from basically an `if` check to `while` change semantics/usage?

oberblastmeister: `?` seems to be non-associative (https://github.com/NixOS/nix/blob/2.10.3/src/libexpr/parser.y#L359). But for some reason `nix repl` accepts (x ? x ? y). Also, you should use `handle_operation_left`

oxalica: > `?` seems to be non-associative

That's because only binary operators with same elements on LHS and RHS can have associativity, in order to dis-ambiguous `(a + b) + c` and `a + (b + c)`.
But `?` has different syntax for LHS and RHS, so `a ? b ? c` can only be `(a ? b) ? c`, not `a ? (b ? c)`. Because `b ? c` is not a AttrPath`.

oberblastmeister:
@oxalica This looks good. I'll merge after it's rebased

src/types.rs
```diff
@@ -407,9 +405,13 @@ typed! [
...
         }
-        /// Return the index
-        pub fn index(&self) -> Option<SyntaxNode> {
-            nth!(self; 1)
+        /// Return the attrpath
+        pub fn key(&self) -> Option<Key> {
+            nth!(self; (Key) 1)
+        }
+        /// Return the or-default branch, if any.
+        pub fn default(&self) -> Option<SyntaxNode> {
```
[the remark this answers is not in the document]
oxalica: It's now `fn default_expr(&self) -> Option<Expr>` since it returns an Expr.

oberblastmeister:
This `or` business is incredibly sus (in the nix language).

oberblastmeister:
It looks like we are parsing `or` wrong. In the nix parser `or` is only an identifier when it is used in an apply expression. In our parser `or` becomes an identifier when it is not used directly after an attrpath in a select expression which is too liberal.

oberblastmeister:
1. This pr already does lots of good improvements, so the issue might not be in the scope of this pr.
2. It might be good to be more liberal, and do a separate pass to warn on using `or` incorrectly. This would be better for error handling because one incorrect `or` won't trigger a bunch of parse errors (because of recovery), and will only trigger one precise dedicated error.

oxalica:
> It looks like we are parsing or wrong.

Yes, it's still not perfect but maybe it's kind of off-topic for this PR.

> This `or` business is incredibly sus (in the nix language).
>  In the nix parser or is only an identifier when it is used in an apply expression. 

No "only". It's also valid in attrpath `let or = 1; in ...`. Not any "apply expression", it's not valid for LHS of Apply.
- `or 1` -> no parse
- `1 or` -> ok

And the "fallback to ident" takes precedence over left-associative of Apply. :exploding_head: 
- `fold or false` -> `(fold or) false`
- `fold false or` -> `foldr (false or)`
- `fold or or` -> no parse

oberblastmeister:
> No "only". It's also valid in attrpath let or = 1; in ....

I was only highlighting the differences, we obviously already parse this correctly

> No "only". It's also valid in attrpath let or = 1; in .... Not any "apply expression", it's only valid for the direct RHS of Apply.

Thanks bro

> And the "fallback to ident" takes precedence over left-associative of Apply.

Yeah unfortunate. That's what happens when you put call inside of inside of `expr_select`, too high precedence!

src/macros.rs
```diff
@@ -7,6 +7,7 @@ macro_rules! T {
     (in)      => ($crate::SyntaxKind::TOKEN_IN);
     (inherit) => ($crate::SyntaxKind::TOKEN_INHERIT);
     (let)     => ($crate::SyntaxKind::TOKEN_LET);
+    (or)      => ($crate::SyntaxKind::TOKEN_OR);
```
oberblastmeister, line 10: Why is this here?

oberblastmeister:
Either `or` needs to be added as a new token in the lexer, or a special case is needed. @oxalica can you do this? see #115

oxalica:
@oberblastmeister
> Either or needs to be added as a new token in the lexer, or a special case is needed.

I created #116 for the fix.

BTW: You are able to push to my PR branch before merging it, if you have write access to the repo, I think?

PR #119, opened 2022-07-27
avoid stack overflow

aaronjanse:
### Summary & Motivation

The fuzzer often finds samples that cause stack overflows, such as a long series of left parenthesis. Fixing this both avoids an error for users and allows the fuzzer to find more interesting bugs in PRs I'm working on.

This may break support for some deeply-nested Nix expressions. I assume that many of these were getting stack overflows before anyway, but maybe not.

### Further context

N/A


aaronjanse:
We might also want to consider looking into the decurse crate

oberblastmeister:
Do we need a test? I think stack size can vary also

aaronjanse:
Stack size definitely varies. Do you think I should remove the test?

oberblastmeister:
Yeah I think so

aaronjanse:
Sounds good. Done in https://github.com/nix-community/rnix-parser/pull/119/commits/b217b2a21251dc664fddd5a7efade45090c0cc4d

PR #120, opened 2022-07-27
improved error handling

aaronjanse:
### Summary & Motivation

This adds helper methods for expecting a token, bumping only if it exists. Using these methods, I improved error handling in two ways:

1. Expressions like `{ foo = 1 } // { bar = 2; }` will now be parsed correctly, with the needed semicolon token simply missing (and appropriate error). Previously, this was parsed as one set with a large error node. 
2. Expressions like `{ foo = bar.; xyz = 1; }` are parsed as missing an attractive path rather than having a large error node consuming most of the tokens. 

### Further context

Replaces #34. This PR is needed to implement autocomplete in rnix-lsp for expressions like `foo.`. I'm sure several other tools will find improved error handling helpful too.


src/parser.rs
```diff
@@ -277,7 +332,14 @@ where
...
         loop {
-            self.parse_attr();
+            match self.peek_any_or_missing(&[TOKEN_INTERPOL_START, TOKEN_STRING_START, TOKEN_IDENT]) {
+                Some(TOKEN_INTERPOL_START) => self.parse_dynamic(),
+                Some(TOKEN_STRING_START) => self.parse_string(),
+                Some(TOKEN_IDENT) => self.expect_ident(),
+                _ => {
+                    break
+                }
+            }
```
aaronjanse, line 342: I personally don't love the repetition here, but I think it's not too bad?

test_data/parser/invalid_syntax/missing-semicolon.nix
```diff
@@ -0,0 +1 @@
+{ foo = 1 bar = 2} // { xyz = 3; }
```
aaronjanse, line 1: It's fun seeing how Github syntax highlighting, on the other hand, is _not_ fault-tolerant

aaronjanse:
I now realize this duplicates work done by @oberblastmeister in #35

PR #126, opened 2022-07-28
Remove Expr Inherit and InheritFrom variants

### LINKED ISSUE #125: LetIn with inherit entry gives wrong node for body
If a `let` expression uses `inherit`, the typed api incorrectly gives the corresponding `NODE_INHERIT` as the body of the `let`.
```rust
use rnix::ast::HasEntry;

fn main() {
    let nix_expr = "let inherit (x) y; in y";
    if let rnix::ast::Expr::LetIn(let_in) =
        rnix::Root::parse(&nix_expr).ok().unwrap().expr().unwrap()
    {
        println!("{:?}", let_in.entries().collect::<Vec<_>>());
        println!("{:?}", let_in.body());
    }
}
```
```
[Inherit(Inherit(NODE_INHERIT@4..18))]
Some(Inherit(Inherit(NODE_INHERIT@4..18)))
```
The cause is the same as the cause of #121: `NODE_INHERIT` is both a valid `Expr` and `Entry`.

In this case, I wonder why is it an `Expr`?

darichey:
### Summary & Motivation
Removes Inherit and InheritFrom variants from the typed Expr enum

Removed `ast::Expr::Inherit` and `ast::Expr::InheritFrom`

### Further context
Fixes #125 

There might be some reason that I'm missing for why these should be `Expr`s.

### DIFF
--- a/src/ast/nodes.rs
+++ b/src/ast/nodes.rs
@@ -172,8 +172,6 @@ node! {
         NODE_ERROR => Error,
         NODE_IF_ELSE => IfElse,
         NODE_SELECT => Select,
-        NODE_INHERIT => Inherit,
-        NODE_INHERIT_FROM => InheritFrom,
         NODE_STRING => Str,
         NODE_PATH => Path,
         NODE_LITERAL => Literal,


aaronjanse:
> There might be some reason that I'm missing for why these should be Exprs.

I'm not sure either. I can't think of a reason for them to be there since afaik NODE_INHERIT{FROM,} can only exist inside within a `let` block or attract set, not as a value?

PR #128, opened 2022-07-28
Allow `TOKEN_OR` when expecting identifier

### LINKED ISSUE #127: `or` is not accepted as an identifier
Following #106 and #116, `or` is no longer accepted as an identifier.

```rust
fn main() {
    let nix_expr = "{ or = 1; }";
    println!(
        "{:#?}",
        rnix::Root::parse(&nix_expr).ok().unwrap()
    );
}
```
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: UnexpectedWanted(TOKEN_OR, 2..11, [TOKEN_IDENT])', src/main.rs:5:43
```

darichey:
### Summary & Motivation
When expecting an identifier in `expect_ident`, allow either `TOKEN_IDENT` OR `TOKEN_OR`.

Also adds a getter for the token in the typed Ident.

### Further context
Fixes #127


oberblastmeister:
This is a good catch! It's not totally correct however, as this allows `or` by itself to be parsed as an expression. You shouldn't change `expect_ident`. You can instead inline it into `parse_attr` and add another case for `TOKEN_OR`.

darichey:
Oh right, oops! Fixed

PR #135, opened 2022-08-06
Refactor parser

oberblastmeister:
- do not check for duplicated identifiers in pattern
- use `T!` macro when parsing
- rename delimiter tokens
- remove `once` argument in favor of separate function


src/tokenizer.rs
```diff
@@ -468,23 +470,3 @@ impl<'a> Iterator for Tokenizer<'a> {
...
-        tokenize(s);
-    }
-
-    #[test]
-    fn test_fuzz() {
-        fuzz("\"");
-        fuzz([39, 39, 34]);
-        fuzz([39, 39, 39, 39, 92]);
-    }
-}
```
darichey, line 490: Why was this removed?

test_data/parser/error/pattern_duplicate_arg.expect
```diff
@@ -1,25 +1,20 @@
-error: argument `a` is duplicated in 8..9
```
darichey, line 1: This test should probably just be removed.

darichey (changes requested):

PR #137, opened 2022-08-16
Re-implement string parts

darichey:
Issues with correctly parsing strings into their interpolation parts still persist after many iterations. This PR makes a drastic move and completely re-implements the functionality by effectively line-for-line copying the reference implementation.

I am not necessarily advocating for merging this change as-is. The code isn't particularly pretty or idiomatic rust-wise, but it has one thing going for it: it correctly parses all strings in nixpkgs.


tazjin:
Without having looked at the code so far, I can say that this does indeed behave correctly. We run the Tvix evaluator against the upstream Nix test suite, which was previously failing on https://github.com/NixOS/nix/blob/master/tests/lang/eval-okay-ind-string.nix#L114-L118 and some other subtleties of that upstream multiline interpolation test.

With this commit we get

    test tests::nix_eval_okay_src_tests_nix_tests_eval_okay_ind_string_nix ... ok

So at least for the edge-cases defined in that file, behaviour seems to be correct! :)

oberblastmeister:
I'll merge once it's updated. We can make the code nicer later

PR #144, opened 2022-09-22
update examples

oberblastmeister:


oberblastmeister:
It is much more useful to pretty-print the ast instead of printing it verbatim.

Issue #146, opened 2022-10-08
Broken paths are silently skipped


tazjin:
Certain path literals are parse errors, most prominently paths with a trailing slash. Compare C++ Nix and Tvix (with rnix-parser):

```
# C++ Nix
nix-repl> ../../
error: path '../../' has a trailing slash

# rnix-parser (through tvix)
tvix-repl> ../../
error[E015]: failed to parse Nix code:
 --> [tvix-repl]:1:1
  |
1 | ../../
  | ^^^^^^ code ended unexpectedly while the parser still expected more
```

These are correctly turned into errors in both cases, however, rnix-parser seems to do something strange if the broken path occurs somewhere in the middle of an AST:

```
# rnix-parser, via tvix, with `--display-ast`
tvix-repl> import ../../
Ident(Ident(NODE_IDENT@0..6))
=> <<primop>> :: lambda

tvix-repl> [ 1 2 ../../ 3 ]
List(List(NODE_LIST@0..16))
=> [ 1 2 3 ] :: list
```

Compare with C++ Nix:

```
nix-repl> import ../../
error: path '../../' has a trailing slash

nix-repl> [ 1 2 ../../ 3 ] 
error: path '../../' has a trailing slash
```

I haven't manually investigated the actual structure produced by rnix-parser in this case yet; it must be in there _somewhere_ (side note: is there a built-in way for some more pretty-printed representation of the AST, either in the crate (that I've missed) or that someone wrote?)

oberblastmeister:
The ast is
```
Root(
    NODE_ROOT@0..14
      NODE_IDENT@0..6
        TOKEN_IDENT@0..6 "import"
      TOKEN_WHITESPACE@6..7 " "
      TOKEN_ERROR@7..13 "../../"
      TOKEN_WHITESPACE@13..14 "\n"
    ,
)
```

This is probably due to https://github.com/nix-community/rnix-parser/blob/f8056c4811b6f5196af64db14b89651e892ad98a/src/kinds.rs#L133
We just bump error tokens as trivia so they appear in the tree, but don't report them as errors.

tazjin:
> We just bump error tokens as trivia so they appear in the tree, but don't report them as errors.

Hmm, not sure I understand the implication of this. We'd have to do a pass over the entire tree and detect any error nodes to get "all" errors? Are there any things other than broken paths that exhibit this behaviour?

[...]

Issue #148, opened 2022-10-15
Weird successfull parsing of invalid path literals


sternenseemann:
### Describe the bug

Was messing around with invalid tilde starting path literals:

* `~foo` is parsed to `Ident(Ident(NODE_IDENT@2..4))`, the `~` seemingly disappears into nowhere. The ident is retained as `oo`, i.e. in tvix you can evaluate `let oo = 42; in ~foo`. This literal should be a parse failure.
* `~foo/bar` is parsed to `Path(Path(NODE_PATH@2..8))`, yielding the path `oo/bar`. This should also be a parse failure.

Nix forbids home path literals that don't start with `~/`, so both expressions are invalid Nix. Probably significant out two chars are dropped from the literals and `~/` is also two chars long.

tazjin:
I'm guessing this is a duplicate of #146. It should be possible to have rnix report these as errors

PR #159, opened 2023-05-18
fix: don't handle TOKEN_ERROR as trivial tokens.

hsjobeki:
### Error Tokens should be seen as something other than trivial. 

This means they are skipped during parsing but must be validated. 

Instead, handle them and fail during parsing. The parser produces an error indicating which token would be expected at that position.

### Summary & Motivation

Details are also in the mentioned issues below.

As far as I can tell, this doesn't break any backward compatibility. rnix should now fail, where it didn't fail on invalid expressions.

@tazjin @sternenseemann: Did I miss something? This shouldn't break anything, right?

### Further context

To ensure correct error handling, I copied the error tests from the tokenizer_tests to the parser_tests. Now the parser should fail and handle the error tokens 

The following issues are fixed:

#146 #148


oberblastmeister:
Thanks for the pr. This should be a sensible change.

PR #163, opened 2024-05-08
Implement reading `SyntaxKind` from strings

abhillman:
### Summary & Motivation

After parsing with `rnix`, I store an intermediate form in JSON which I wish to be able to re-serialize into an AST. Allowing reading `SyntaxKind` from strings is helpful for this use-case.

n/a

### Further context

n/a


philiptaron (approved):
No new deps needed, just a simple `FromStr` impl. Nice.

PR #167, opened 2024-12-13
Added support for the experimental pipe operators

aftix:
### Summary & Motivation

Adds support for the pipe operator as described in https://github.com/NixOS/rfcs/pull/148 . It is available with on Nix 2.24 with the pipe-operators experimental feature and on Lix with the pipe-operator experimental feature. The operator precedence is between function application and attribute selection.

This should not effect parsing of any input text that already parsed without error before this change: it only adds more valid input.

### Further context

This library is used downstream in places like statix and alejandra. Currently, using pipe operator syntax in your nix project makes any tools downstream from this library completely unusable. Issues https://github.com/oppiliappan/statix/issues/88 and https://github.com/kamadorueda/alejandra/issues/436 are blocked on this library not supporting the operators. 

Since this doesn't change parsing of nix code that doesn't use the pipe operator, it would be a much better user experience to have the downstream tools error with "experimental feature pipe-operator not enabled" or something when the feature is off (if they wish to) than to have the tools plain not work with any codebase using pipe operators.


oberblastmeister:
Hi, thanks for the pr! Could you rewrite the tests as snapshot tests and add them to the `test_data` directory? Thanks

src/parser.rs
```diff
@@ -583,8 +583,11 @@ where
 
         checkpoint
     }
+    fn parse_pipe(&mut self) -> Checkpoint {
+        self.parse_left_assoc(Self::parse_simple, T!["|>"] | T!["<|"])
```
oberblastmeister, line 587: shouldn't these have different associativity?

aftix: You're right - it wasn't explicitly mentioned in the RFC but using nix repl i see that <| is right associative

aftix:
> Hi, thanks for the pr! Could you rewrite the tests as snapshot tests and add them to the `test_data` directory? Thanks

I've moved the tests to test_data with tests for associativity / the correct precedence between |> and <|

oberblastmeister:
I think the precedence is wrong https://nix.dev/manual/nix/2.25/language/operators#pipe-operators. It should be the operator with the lowest binding power. You can replace `parse_math` with a new function like `parse_pipe`. `parse_math` is the outermost parsing function and thus has the lowest binding power. Then can you add more tests combining pipes with other operators like `+`?

aftix:
yeah you're right, i've changed the precedence and added tests for the precedence in relation to other operators

oberblastmeister:
looks good to me

mightyiam:
Thank you!

PR #168, opened 2024-12-19
Bumped library version

aftix:
I realized my previous PR needed a semver bump, since it added variants to a public enum which is a breaking change (users' match statements could become non-exhaustive and error).


RobWalt:
@oberblastmeister If you find the time, could you also do a release on crates.io? We kind of need the latest changes downstream in statix.

oberblastmeister:
@RobWalt The new version has been released.

Issue #166, opened 2024-11-03
Distinguish between HPATH, SPATH and plain PATH


sternenseemann:
### Description

Paths are currently just parsed into parts of strings and identifiers (in case of interpolation). Nix has three distinct types of paths that need to be treated differently (at least they are in evaluation):

- Regular paths may be subject to path interpolation. They also need may need to be resolved to absolute paths. We could be very precise and distinguish between absolute and relative paths (Upstream C++ Nix doesn't), but probably path libraries are in general well equipped to correctly to consume both relative and absolute paths and ensure that they are absolute…
- Home relative paths may be subject to path interpolation. They need to be resolved relative to the current user's home directory.
- NIX_PATH expressions (SPATH tokens) need to be resolved based on the configured nix search path (usually via the `NIX_PATH` environment variable). They are not subject to path interpolation.

It would be great if they were distinct on the type level. Currently, you need to parse what `rnix` gives you again to figure out what kind of path you are dealing with.

PR #172, opened 2025-04-20
Distinguish different types of Nix paths at the type level

domenkozar:
https://github.com/nix-community/rnix-parser/issues/166 cc @sterni


domenkozar:
Any chance for a review?

oberblastmeister:
Sorry for leaving this. I think it would be better to create a single `NODE_PATH` that is the parent of all `TOKEN_PATH_*` tokens, similar to how the `NODE_LITERAL` node is the parent of `TOKEN_INTEGER`, `TOKEN_FLOAT`, etc.

domenkozar:
cc @sternenseemann

oberblastmeister:
I think we can keep it as is. Maybe we can create an enum for the path nodes using the `node!` macro similar to `Expr`?

domenkozar:
@oberblastmeister thanks, I've done that and implement a hierarchy of traits on top of it to extract parts of the path out :)

domenkozar:
ping

src/ast/path_util.rs
```diff
@@ -1,65 +1,299 @@
...
+
+impl PathNode for PathAbs {}
+impl PathNode for PathRel {}
+impl PathNode for PathHome {}
+impl PathNode for PathSearch {}
+impl PathNode for Path {}
+
+/// A path part - can be literal or interpolation
+#[derive(Debug, Clone)]
+pub enum PathPart {
```
oberblastmeister, line 23: how about we `InterpolPart<PathContent>`

oberblastmeister:
Is there a reason why we need the `PathNode` trait?

domenkozar:
@oberblastmeister ping :)

domenkozar:
Ping @oberblastmeister

PR #179, opened 2026-01-29
feat: add Root::parse_bytes for non-UTF8 input

### LINKED ISSUE #173: Support parsing non-strings
### Description
Nixcpp doesn't fail to parse `.nix` files that contain non-UTF8 characters, however rnix doesn't allow this (as you have to pass in a `&str`, and you can't construct this from a non-UTF8 string)

Ideally we could pass any `AsRef<[u8]>`. Things like the `bstr` crate make this less annoying to interact with.

### Considered alternatives
Explicitly documenting why it's a non-goal.

### Additional context
https://git.snix.dev/snix/snix/issues/124

philiptaron:
## Summary

- Add `Root::parse_bytes(&[u8])` method that handles non-UTF8 input
- Invalid UTF-8 sequences are replaced with U+FFFD (replacement character)

## Background

The C++ Nix parser uses [`%option 8bit`](https://github.com/NixOS/nix/blob/master/src/libexpr/lexer.l#L1) in its flex lexer, which allows it to handle arbitrary byte values without UTF-8 validation. The raw bytes are preserved in string literals.

rnix uses [Rowan](https://github.com/rust-analyzer/rowan) for its syntax tree, which requires valid UTF-8. This means we cannot preserve arbitrary bytes exactly. Instead, `parse_bytes` does lossy UTF-8 conversion - invalid sequences become U+FFFD (�).

**Behavior comparison:**
| Input bytes | nix output | rnix parse_bytes output |
|-------------|------------|-------------------------|
| `{ x = "\xff"; }` | `{ x = "\xff"; }` (raw 0xFF preserved) | `{ x = "�"; }` (U+FFFD replacement) |

This is sufficient for most use cases (linting, formatting, analysis) where exact byte preservation isn't required.

## Example

```rust
// { x = "\xff"; } with raw 0xFF byte (invalid UTF-8)
let bytes: &[u8] = &[0x7b, 0x20, 0x78, 0x20, 0x3d, 0x20, 0x22, 0xff, 0x22, 0x3b, 0x20, 0x7d];
let parse = Root::parse_bytes(bytes);
assert!(parse.errors().is_empty());
// Output text: { x = "�"; }
```

## Test plan

- [x] Added test case `non_utf8_can_be_parsed_with_parse_bytes_issue173`
- [x] All existing tests pass

Closes #173

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -44,9 +44,27 @@ pub type SyntaxNodeChildren = rowan::SyntaxNodeChildren<NixLanguage>;
 pub use ast::Root;
 
 impl Root {
+    /// Parse a string into an AST.
     pub fn parse(s: &str) -> Parse<Root> {
         let (green, errors) = parser::parse(tokenize(s));
         Parse { green, errors, _ty: PhantomData }
+    }
+
+    /// Parse bytes into an AST.
+    ///
+    /// If the input is valid UTF-8, it is used directly. Otherwise, invalid
+    /// UTF-8 sequences are replaced with U+FFFD (replacement character).
+    ///
+    /// This allows parsing `.nix` files that contain non-UTF-8 bytes, matching
+    /// the behavior of the C++ Nix parser.
+    pub fn parse_bytes(bytes: &[u8]) -> Parse<Root> {
+        match std::str::from_utf8(bytes) {
+            Ok(s) => Self::parse(s),
+            Err(_) => {
+                let s = String::from_utf8_lossy(bytes);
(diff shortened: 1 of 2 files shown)


flokli:
I'm honestly not sure this is a great idea.

We would silently (!) parse things differently than what the user has requested.

I'll explore the field a bit more, and see how this is observable from nixlang.

philiptaron:
> I'm honestly not sure this is a great idea.

Me either. I've left it here as a "hmmm."

PR #181, opened 2026-02-08
Add `__curPos` node

3timeslazy:
### Summary & Motivation

Add support for __curPos symbol.

Its behavior is not described in the nix manual, but there is a reference in Lix: https://docs.lix.systems/manual/lix/nightly/language/constructs.html#keywords-__curPos

As far as I call tell without being familiar with CppNix codebase, having __curPos as an AST kind matches how other parses handle that:
- https://github.com/oxalica/nil/blob/504599f7e555a249d6754698473124018b80d121/crates/ide/src/ty/infer.rs#L337-L350
- https://github.com/NixOS/nix/blob/3b473c4be5e6abc58aa43271f35ff127af806c55/src/libexpr/parser.y#L315-L320

In certain contexts, nix expressions containing `__curPos` used to be represented as `NODE_IDENT`, but with this change they will be of type `NODE_CUR_POS`

### Further context

This change is part of __curPos implementation in snix: https://cl.snix.dev/c/snix/+/30959/1


philiptaron (changes requested):
Thanks for this PR! The idea of giving `__curPos` its own AST node type seems the right direction -- however, making `__curPos` a distinct token at the **lexer** level (`TOKEN_CUR_POS`) introduces regressions in contexts where it should still behave as a regular identifier. 

In Nix, `__curPos` is lexed as a normal `ID` — it only gets special semantics in expression position ([parser.y#L315-L320](https://github.com/NixOS/nix/blob/3b473c4be5e6abc58aa43271f35ff127af806c55/src/libexpr/parser.y#L315-L320)).

I've pushed regression tests to the [`cur-pos-regression-tests`](https://github.com/nix-community/rnix-parser/tree/cur-pos-regression-tests) branch (commit 9609a7e) that demonstrate the failures. These are all valid Nix expressions that parse correctly on `master` today but break with this PR:

| Test | Expression | Problem |
|---|---|---|
| `cur_pos-lambda.nix` | `__curPos: __curPos` | Lambda param — the `TOKEN_IDENT` arm in `parse_simple` handles `:` → lambda, but the new `T![__curPos]` arm doesn't |
| `cur_pos-pattern-bind.nix` | `__curPos @ { }: __curPos` | Pattern bind — same issue, `@` → pattern bind only checked in `TOKEN_IDENT` arm |
| `cur_pos-pattern-field.nix` | `{ __curPos, ... }: __curPos` | Pattern field — `parse_pattern` only accepts `TOKEN_IDENT`; also the `{` lookahead in `parse_simple` doesn't match `TOKEN_CUR_POS` so this is misclassified as a set |
| `cur_pos-pattern-field-default.nix` | `{ __curPos ? 1 }: __curPos` | Pattern field with default — same lookahead + pattern issue |

The root cause is that giving `__curPos` its own token kind means every place that accepts `TOKEN_IDENT` needs updating (pattern parsing, lambda detection, set-vs-pattern lookahead, etc.), which is fragile and error-prone.

Maybe keep `__curPos` as `TOKEN_IDENT` at the tokenizer level (matching Nix) and only distinguish it at the parser level — in `parse_simple`'s expression position, check if the identifier text is `"__curPos"` and wrap it in `NODE_CUR_POS` instead of `NODE_IDENT`. This way all identifier contexts (lambdas, patterns, attributes) continue working unchanged.

A couple of other things I noticed:
- The `FromStr` impl on `SyntaxKind` in `kinds.rs` is missing entries for `TOKEN_CUR_POS` and `NODE_CUR_POS`
- The catch-all error in `parse_simple` doesn't list `TOKEN_CUR_POS` in its "wanted" tokens

3timeslazy:
Hey @philiptaron,

I looked into your comments and made another __curPos draft.

Some reasoning behind it:

As I mentioned in https://github.com/nix-community/rnix-parser/pull/181#issuecomment-3900266196, I think the most convenient API for
library clients would be to return `NODE_CUR_POS + TOKEN_CUR_POS` where appropriate.

First I tried something like that

```rust
let ident = self.peek_raw().map(|&(_, s)| s);
let checkpoint = self.checkpoint();
self.expect_ident();

match self.peek() {
    // lambda branch
    // pattern bind branch
}

if ident == "__curPos" {
    self.start_node_at(checkpoint);
    self.finish_node();
}
```

But this code results in

```
NODE_CUR_POS
    NODE_IDENT
        TOKEN_IDENT
```

Because of the `self.expect_ident` call in the beginning which produces a NODE_IDENT.

So I implemented a little look ahead method that reads tokens until there is N non-trivia tokens but without side effects of other peek methods that create nodes

philiptaron (approved):

philiptaron:
I've released 0.14.0 with this change and a bug fix for absolute paths (#182)

PR #186, opened 2026-06-01
Fix mistakenly allowing empty dynamic selects

Starnick4444:
### Summary & Motivation

The parser allows empty dynamic selects (e.g. `{ foo = 1; }.${}`, which should be an error.


Starnick4444:
@philiptaron could you take a look please?

philiptaron (approved):
Validated this matches upstream Nix.

